PR #15, opened 2022-11-22
Okim/user identifiers

oliviakim321:
- add user identity class to store user identifiers map
- add input validation to make sure the input is of the correct shape
- rename app user id --> client user id
- update readme (will take another pass later)


Creative/ATTNSDK.m
```diff
@@ -13,37 +13,34 @@ @implementation ATTNSDK {
...
 - (id)initWithDomain:(NSString *)domain mode:(NSString *)mode {
     _domain = domain;
     _mode = mode;
+    _userIdentity = [[ATTNUserIdentity alloc] init];
     return [super init];
 }
 
-- (void)identify: (NSString *)appUserId {
-    _appUserId = appUserId;
+- (void)identify:(NSDictionary *)userIdentifiers {
```
wyattjoe40, line 30: We will accept NSStrings here for back-compat, right? If so, doesn't have to be in this PR, just confirming

Creative/ATTNUserIdentity.m
```diff
@@ -0,0 +1,60 @@
...
+    [self validateIdentifiers:identifiers];
+    _identifiers = identifiers;
+    
+    return self;
+}
+
+- (void)mergeIdentifiers:(nonnull NSDictionary *) newIdentifiers {
+    [self validateIdentifiers:newIdentifiers];
+
+    NSMutableDictionary *currentIdentifiersCopy = [_identifiers mutableCopy];
```
wyattjoe40, line 45: why do we copy the identifiers dictionary?

wyattjoe40 (approved):

Creative/ATTNUserIdentity.m
oliviakim321: NSDictionary is immutable so we can't add to it - we create a mutable copy that we can add the new identifiers to

Creative/ATTNSDK.m
oliviakim321: yup! was going to add that in a new PR but ill actually just add here

Creative/ATTNUserIdentity.m
wyattjoe40: got it, thanks!

PR #17, opened 2022-11-24
Improve iOS project structure and enable testing

wyattjoe40:
Creates a new directory structure:
* Sources - source code of the SDK
* Tests - test source code of the SDK
* Example - example app
* Framework - additional files used to build the SDK into a dynamic framework

The Example app project has been updated:
* No longer requires a code change to switch between using the attentive-ios-sdk Pod and the local source. Now you can select a scheme and it will build the correct one:
* Scheme `Example - Local` - builds the Example app with the local source
* Scheme `Example - Pod` - builds the Example app with the Pod source

A new XCode workspace is added: `attentive-ios-sdk.xcworkspace` . This workspace contains these projects:
* attentive-ios-sdk - the SDK
* Pods - the CocoaPod-created project that builds the published attentive-ios-sdk Pod
* Example - the example app. This project depends on the other two.


oliviakim321 (approved):
💯

PR #22, opened 2022-11-29
add identifiers to creatives call

oliviakim321:
- add identifiers to creatives call
- isEqual --> isEqualToString
- fix success log in debug mode

### DIFF
--- a/Sources/ATTNSDK.m
+++ b/Sources/ATTNSDK.m
@@ -82,10 +82,7 @@
 
 - (nonnull NSString *)buildCompanyCreativeUrl {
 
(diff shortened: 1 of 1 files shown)


Sources/ATTNSDK.m
```diff
@@ -79,6 +72,72 @@ - (void)trigger:(UIView *)theView {
     }
 }
 
+- (nonnull NSString *)buildCompanyCreativeUrl {
+
+    NSURLComponents *components = [[NSURLComponents alloc] init];
```
wyattjoe40, line 77: nit: you can use `initWithString` and pass it the URL. might be easier to read

wyattjoe40 (approved):

PR #26, opened 2022-12-05
A few random fixes

wyattjoe40:
* Adds button to clear the user
* Adds visitorId to identify API call
* Returns actual error when geo-adjusted domain cannot be found


Sources/ATTNAPI.m
```diff
@@ -31,6 +31,7 @@ - (id)init {
 }
 
 - (void)sendUserIdentity:(ATTNUserIdentity *)userIdentity domain:(NSString *)domain {
+    NSLog(@"visitorId: %@", userIdentity.visitorId);
```
oliviakim321, line 34: should we remove this?

oliviakim321 (approved):

Sources/ATTNAPI.m
wyattjoe40: @oliviakim321 removed!

PR #30, opened 2022-12-19
Adds EventTracker and Purchase events

wyattjoe40:
* Adds an `ATTNEventTracker`. This class is a singleton so it's easily accessible anywhere in the app. This class is used to send events to Attentive.
* Adds OCMock library for testing
* Updates Example app with a button to send a Purchase


Tests/ATTNSDKTest.m
```diff
@@ -33,7 +33,7 @@ - (void)tearDown {
 - (void)testExample {
     ATTNUserIdentity* identity = [[ATTNUserIdentity alloc] init];
     ATTNAPI* api = [[ATTNAPI alloc] init];
-    ATTNSDK* sdk = [[ATTNSDK alloc] initWithDomain:@"domainValue" mode:@"Production" userIdentity:identity api:api];
+    // ATTNSDK* sdk = [[ATTNSDK alloc] initWithDomain:@"domainValue" mode:@"Production" userIdentity:identity api:api];
```
Mackey22, line 36: should we remove this file altogether?

Mackey22 (commented):
Lgtm. what are the ramifications of adding OCMock? Any build issues that could come up?

wyattjoe40:
> Lgtm. what are the ramifications of adding OCMock? Any build issues that could come up?

I added OCMock to just the Unit Test build/product, not the actual SDK library, so I don't think there are many ramifications, and if they are they should be isolated to the Unit Test build/product.

Tests/ATTNSDKTest.m
wyattjoe40: yeah good call, removed

Mackey22 (approved):

PR #31, opened 2022-12-20
Sends Purchase events to server

wyattjoe40:
* Hooks up `ATTNEventTracker` to the `ATTNAPI`
* Updates `ATTNAPI` to send PurchaseEvents to the server
* Adds `nullable` annotations to some properties


Sources/ATTNAPI.m
```diff
@@ -42,6 +95,95 @@ - (void)sendUserIdentity:(ATTNUserIdentity *)userIdentity domain:(NSString *)dom
...
+        }
+        
+        NSLog(@"%@", [NSString stringWithFormat:@"Successfully sent event of type '%@'", request.eventNameAbbreviation]);
+    }];
+    
+    [task resume];
+}
+
+// A single event can create multiple requests to our servers. This method converts an event to multiple request objects.
+- (NSArray<EventRequest *>*)convertEventToRequests:(id<ATTNEvent>)event {
```
Mackey22, line 141: we're gonna want an OrderConfirmed event produced as well

Sources/ATTNAPI.m
```diff
@@ -42,6 +95,95 @@ - (void)sendUserIdentity:(ATTNUserIdentity *)userIdentity domain:(NSString *)dom
     }];
 }
 
+- (void)sendEvent:(id<ATTNEvent>)event userIdentity:(ATTNUserIdentity*)userIdentity domain:(NSString*) domain {
+    [self getGeoAdjustedDomain:domain completionHandler:^(NSString* geoAdjustedDomain, NSError* error) {
```
Mackey22, line 99: I said it above but repeating here just to make sure. I think if possible we want to hold domain in memory to not need it a second time

Sources/ATTNAPI.m
```diff
@@ -42,6 +95,95 @@ - (void)sendUserIdentity:(ATTNUserIdentity *)userIdentity domain:(NSString *)dom
...
+        [self sendEventInternal:event userIdentity:userIdentity domain:domain];
+    }];
+}
+
+- (void)sendEventInternal:(id<ATTNEvent>)event userIdentity:(ATTNUserIdentity*)userIdentity domain:(NSString*) domain {
+    // slice up the Event into individual EventRequests
+    NSArray<EventRequest *>* requests = [self convertEventToRequests:event];
+    
+    for (EventRequest* request in requests) {
+        [self sendEventInternalForRequest:request userIdentity:userIdentity domain:domain];
```
Mackey22, line 114: does each one of these requests open the radio? should we be working on reducing the number of requests rn or you think it's fine?

Sources/ATTNAPI.h
```diff
@@ -11,15 +11,18 @@
...
 NS_ASSUME_NONNULL_BEGIN
 
 @interface ATTNAPI : NSObject
 
-- (id)init;
+- (instancetype)init;
 
 - (void)sendUserIdentity:(ATTNUserIdentity *) userIdentity domain:(NSString *) domain;
 
+- (void)sendEvent:(id<ATTNEvent>)event userIdentity:(ATTNUserIdentity*)userIdentity domain:(NSString*) domain;
```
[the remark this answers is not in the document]
Mackey22: I think this might actually be properly accounted for given what i saw way lower in the PR but lmk if i'm wrong

Events/ATTNItem.m
```diff
@@ -16,6 +16,7 @@ - (instancetype)initWithProductId:(NSString*) productId productVariantId:(NSStri
         self->_productId = productId;
         self->_productVariantId = productVariantId;
         self->_price = price;
+        self->_quantity = 1;
```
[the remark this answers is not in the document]
wyattjoe40: ATTNItem* item = ...;
item.quantity = 5;

Sources/ATTNAPI.m
wyattjoe40: Do we want that now?

Sources/ATTNAPI.m
wyattjoe40: Yeah, domain makes sense to me. That shouldn't change after startup. I'll add to the constructor.

Sources/ATTNAPI.m
wyattjoe40: Honestly I think events are so infrequent right now that it won't matter. The radio should stay on between requests since they're back to back, and the radio may even still be on from the app's other communications with their own servers.

Sources/ATTNAPI.m
wyattjoe40: And in the future we'll store the geoAdjustedDomain as a member variable as well

Sources/ATTNAPI.h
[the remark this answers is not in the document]
wyattjoe40: We talked about this offline. Passing in a UserIdentity allows us better multi-threading safety. We can "snapshot" the user identifiers before passing to the ATTNAPI

PR #37, opened 2023-01-10
Opens Privacy/Terms links in the browser app

wyattjoe40:


Sources/ATTNSDK.m
```diff
@@ -206,6 +206,15 @@ - (void)webView:(WKWebView *)webView decidePolicyForNavigationAction:(WKNavigati
         [UIApplication.sharedApplication openURL:url];
         decisionHandler(WKNavigationActionPolicyCancel);
         return;
+    } else if ([navigationAction.request.URL.scheme isEqualToString:@"http"] || [navigationAction.request.URL.scheme isEqualToString:@"https"]) {
```
oliviakim321, line 209: should we make this a case insensitive check like we do for android?

oliviakim321 (approved):

Sources/ATTNSDK.m
wyattjoe40: Done!

oliviakim321 (approved):

PR #38, opened 2023-01-17
AddToCart and ProductView and Example app update

wyattjoe40:
Library:
* Adds AddToCart and ProductView events
* For better organization, adds a new "Events" Group in Xcode and moves the event classes to it. (note: this does not move the files on disk; they are still in the "Sources" directory on disk)

Example app:
* Adds a new NavigationView that allows for multiple ViewControllers
* Adds a new UI for a product page:
Screen Shot 2023-01-17 at 3 09 53 PM


Sources/ATTNAPI.m
```diff
@@ -189,6 +192,38 @@ - (void)sendEventInternalForRequest:(EventRequest*)request userIdentity: (ATTNUs
...
+        if ([addToCart.items count] == 0) {
+            NSLog(@"No items found in the AddToCart event.");
+            return @[];
+        }
+        
+        for (ATTNItem* item in addToCart.items) {
+            NSMutableDictionary* metadata = [[NSMutableDictionary alloc] init];
+            [self addProductDataToDictionary:item dictionary:metadata];
+            
+            [eventRequests addObject:[[EventRequest alloc] initWithMetadata:metadata eventNameAbbreviation:@"c"]];
```
oliviakim321, line 208: nit: should we make these abbreviations constants like we do for the user identity keys?

wyattjoe40: good call, literal strings are lazy. updated!

PR #39, opened 2023-01-20
Caches geoadjusted domain

wyattjoe40:
* Caches the geoadjusted domain (GAD) so that subsequent calls don't have to call our API again to get the GAD
* This implementation is simple and more optimized than the current code, but it is not optimal. Why is it not optimal? There can still be multiple calls to get the GAD. Here's how: Assume the GAD is not cached. Event 1 is being sent by the SDK and it triggers an async call to get the GAD. Before that call finishes Event 2 is then sent by the SDK - this triggers another async call to get the GAD. Each of these calls will complete, in some unknown order, and each will overwrite the cachedGeoAdjustedDomain. We should not get different GAD values from each call, so the cached GAD will end up being the same.

### DIFF
--- a/Tests/ATTNAPITest.m
+++ b/Tests/ATTNAPITest.m
@@ -18,6 +18,7 @@
 #import "ATTNCart.h"
 
 static NSString* const TEST_DOMAIN = @"some-domain";
+static NSString* const TEST_GEO_ADJUSTED_DOMAIN = @"some-domain";
 
 @interface ATTNAPI (Testing)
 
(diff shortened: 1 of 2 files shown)


Tests/ATTNAPITest.m
```diff
@@ -18,6 +18,7 @@
 #import "ATTNCart.h"
 
 static NSString* const TEST_DOMAIN = @"some-domain";
+static NSString* const TEST_GEO_ADJUSTED_DOMAIN = @"some-domain";
```
oliviakim321, line 21: nit: should we name this domain to `@"some-domain-ca";` or something just to not confuse why we have two of the same variables in this class?

oliviakim321 (approved):

Tests/ATTNAPITest.m
wyattjoe40: 😵‍💫 yes, it should be different

wyattjoe40: updated!

PR #43, opened 2023-02-11
Adds a Swift example app

wyattjoe40:
* Adds another example app written in Swift
* The Swift app does not use scenes (like our objective-c Example app) because our Swift customers so far don't use scenes (i.e. UISceneDelegate)


ExampleSwift/ExampleSwift/ViewController.swift
```diff
@@ -0,0 +1,44 @@
...
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        self.view.backgroundColor = .systemGray3
+    }
+    
+    @IBAction func creativeBtnPressed(sender: Any) {
+        self.clearCookies()
+        
+        self.getAttentiveSdk().trigger(self.view)
+    }
```
oliviakim321, line 25: nit: should we add the comments that we have in the other example to this file?

oliviakim321 (dismissed):

oliviakim321 (approved):

PR #44, opened 2023-02-17
add uitests

oliviakim321:


Example/Example/ViewController.m
```diff
@@ -66,4 +75,18 @@ - (IBAction)clearUserButtonPressed:(id)sender {
...
 
+// Method for setting up UI Tests. Only used for testing purposes
+- (void)setupForUITests {
+    // Override the hard-coded domain & mode with the values from the environment variables
+    _domain = [[[NSProcessInfo processInfo] environment] objectForKey:@"com.attentive.Example.DOMAIN"];
+    _mode = [[[NSProcessInfo processInfo] environment] objectForKey:@"com.attentive.Example.MODE"];
+
+    // Reset the standard user defaults - this must be done from within the app to avoid
+    // race conditions
+    NSString * persistentDomainToRemove = [[[NSProcessInfo processInfo] environment] objectForKey:@"com.attentive.Example.PERSISTENT_DOMAIN_TO_REMOVE"];
```
wyattjoe40, line 86: Perhaps using [synchronize](https://developer.apple.com/documentation/foundation/nsuserdefaults/1414005-synchronize) would fix the race condition? I'm not sure, but seems likely

wyattjoe40 (approved):
One comment. If it doesn't work feel free to ignore.

Example/Example/ViewController.m
oliviakim321: hmm did think about that but opted not to since the docs say "this method is unnecessary and shouldn't be used"

wyattjoe40: For general production use, yeah, agreed, but for tests I think it's acceptable. Since this code works I'm cool leaving it as-is, but figured I'd mention.

PR #45, opened 2023-02-24
Update UI Tests for AWS Device Farm

oliviakim321:
- tap 10% OFF instead of UNLOCk since it's more in the center of the app & is less likely to be cut off
- only verify the sms app is opened locally since aws device farm doesn't allow use of sms apps
- update env variables names to respect env variable naming conventions
- fix setupUITest function to only override domain & mode when they are set

### DIFF
--- a/Example/Example/ViewController.m
+++ b/Example/Example/ViewController.m
@@ -78,8 +78,8 @@ ATTNSDK *sdk;
 // Method for setting up UI Tests. Only used for testing purposes
 - (void)setupForUITests {
     // Override the hard-coded domain & mode with the values from the environment variables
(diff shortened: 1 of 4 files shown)


Example/Example/ViewController.m
```diff
@@ -21,7 +21,7 @@ - (void)viewDidLoad {
     self.view.backgroundColor = [UIColor systemGray3Color];
     
     // Replace with your Attentive domain to test with your Attentive account
-    _domain = @"YOUR_ATTENTIVE_DOMAIN";
+    _domain = @"mobileapps";
```
wyattjoe40, line 24: Mistake?

wyattjoe40 (approved):

Example/Example/ViewController.m
oliviakim321: whoops yup ty!

PR #49, opened 2023-03-14
Adds SDK Version to user-agent

wyattjoe40:
Main purpose:
* Adds a user agent to all HTTP requests. Example: `Example---Local/1.0.1 (iPhone; iOS Version 16.2 (Build 20C52)) attentive-ios-sdk/0.3.2`

Additional updates:

* Shows the `domain` in the Example app UI now
* Updates the name of the local `attentive-ios-sdk` framework so there are no collisions with the Pod framework name. The issue is they both want to use the name `attentive-ios-sdk`, but only one can. I changed the local framework name to `attentive-ios-sdk-framework` since it's only being used by our Example apps.
* Now that the name collision is fixed, I changed back to `use_framework` for the Pods. This creates a dynamic framework instead of a static framework, which is likely what our customers will use. The deleted files in this PR are due to the removal of the static framework.
* Updated the SDK version that the Example apps rely on


Internal/ATTNAppInfo.h
```diff
@@ -0,0 +1,31 @@
+//
```
wyattjoe40, line 1: @oliviakim321 The real changes start here

Example/Example/ViewController.m
```diff
@@ -43,6 +45,11 @@ - (void)viewDidLoad {
                           IDENTIFIER_TYPE_KLAVIYO_ID: @"555555",
                           IDENTIFIER_TYPE_CUSTOM_IDENTIFIERS: @{@"customId": @"customIdValue"}
     };
+    [sdk identify:_userIdentifiers];
+    
+    
+    // Random Example app stuff
```
oliviakim321, line 51: nit: should we clarify this comment a little?

oliviakim321 (approved):

Example/Example/ViewController.m
wyattjoe40: you're right, this wasn't clear. Updated to a hopefully more clear comment

PR #53, opened 2023-03-21
Adds SDK version

wyattjoe40:
- Adds SDK version tracking
- Sends INFO events
- Changes events endpoint from GET to POST requests


ATTNTestEventUtils.h
```diff
@@ -18,6 +18,7 @@
 #import "ATTNOrder.h"
 #import "ATTNPrice.h"
 #import "ATTNCart.h"
+#import "ATTNInfoEvent.h"
```
wyattjoe40, line 21: This class needs to be moved to the `Tests` directory. Will do this in a separate PR.

PR #65, opened 2023-05-25
TRIGINTG-600 Improve identify to allow empty strings, but log about it

mrhota:


Sources/ATTNParameterValidation.m
```diff
@@ -31,8 +39,12 @@ + (void)verifyNotNil:(nonnull NSObject *)inputValue inputName:(nonnull const NSS
...
 
 + (void)verifyStringOrNil:(nullable NSString *)inputValue inputName:(nonnull const NSString *)inputName {
-  if ([ATTNParameterValidation isNotNil:inputValue] && ![ATTNParameterValidation isStringAndNotEmpty:inputValue]) {
-    [NSException raise:@"Bad Identifier" format:@"%@ should be a non-empty NSString", inputName];
+  if ([ATTNParameterValidation isNotNil:inputValue]) {
+    if (![ATTNParameterValidation isString:inputValue]) {
+      [NSException raise:@"Bad Identifier" format:@"%@ should be NSString", inputName];
+    } else if ([ATTNParameterValidation isEmpty:inputValue]) {
+      NSLog(@"Identifier %@ should be non-empty", inputName);
+    }
```
mrhota, line 47: Now that I think about it, I want a separate method instead of rewriting this one.

Sources/ATTNParameterValidation.m
```diff
@@ -36,6 +44,18 @@ + (void)verifyStringOrNil:(nullable NSString *)inputValue inputName:(nonnull con
   }
 }
 
++ (void)verifyString:(nullable NSString *)inputValue inputName:(nonnull const NSString *)inputName {
+  if ([ATTNParameterValidation isNil:inputValue]) {
+    return;
+  }
+  if (![ATTNParameterValidation isString:inputValue]) {
+    [NSException raise:@"Bad Identifier" format:@"%@ should be NSString", inputName];
+  }
```
oliviakim321, line 53: I think we shouldn't be using `NSException raise` because that can crash clients' apps - let's just log an error here instead

Sources/ATTNUserIdentity.m
```diff
@@ -53,11 +53,14 @@ - (void)mergeIdentifiers:(nonnull NSDictionary *)newIdentifiers {
...
-  [ATTNParameterValidation verifyStringOrNil:identifiers[IDENTIFIER_TYPE_SHOPIFY_ID] inputName:IDENTIFIER_TYPE_SHOPIFY_ID];
-  [ATTNParameterValidation verifyStringOrNil:identifiers[IDENTIFIER_TYPE_KLAVIYO_ID] inputName:IDENTIFIER_TYPE_KLAVIYO_ID];
+  NSArray *keys = @[ IDENTIFIER_TYPE_CLIENT_USER_ID,
+                     IDENTIFIER_TYPE_PHONE,
+                     IDENTIFIER_TYPE_EMAIL,
+                     IDENTIFIER_TYPE_SHOPIFY_ID,
+                     IDENTIFIER_TYPE_KLAVIYO_ID ];
+  [keys enumerateObjectsUsingBlock:^(NSString *key, NSUInteger idx, BOOL *stop) {
+    [ATTNParameterValidation verifyString:identifiers[key] inputName:key];
+  }];
```
oliviakim321, line 63: could this part just be 
```
for (id object in array) {
    // do something with object
}
```
?

Sources/ATTNParameterValidation.m
```diff
@@ -36,6 +44,18 @@ + (void)verifyStringOrNil:(nullable NSString *)inputValue inputName:(nonnull con
   }
 }
 
++ (void)verifyString:(nullable NSString *)inputValue inputName:(nonnull const NSString *)inputName {
```
oliviakim321, line 47: can we just add this code to `verifyStringOrNil` instead of creating a new method?

Sources/ATTNUserIdentity.m
mrhota: well that certain seems nicer 😄

Sources/ATTNParameterValidation.m
oliviakim321: yup, this is an internal class so you can change the existing code! and you can fix/add tests to verify as well

PR #66, opened 2023-06-01
No exceptions; only logs

mrhota:
Instead of raising exceptions (which are likely to kill an unsuspecting app), we'll just log an error and (usually) return.

In a future PR, we will move to the newer Unified Logging system.

### DIFF
--- a/Sources/ATTNCustomEvent.m
+++ b/Sources/ATTNCustomEvent.m
@@ -18,6 +18,7 @@
     NSString *invalidCharInType = [ATTNCustomEvent findInvalidCharacterInType:type];
     if (invalidCharInType != nil) {
       NSLog(@"Invalid character '%@' in CustomEvent type '%@'", invalidCharInType, type);
+      return nil;
     }
 
     [ATTNParameterValidation verifyNotNil:properties inputName:@"properties"];
@@ -27,6 +28,7 @@
       NSString *invalidCharInKey = [ATTNCustomEvent findInvalidCharacterInPropertiesKey:key];
       if (invalidCharInKey != nil) {
         NSLog(@"ERROR: Invalid character '%@' in CustomEvent property key '%@'", invalidCharInKey, key);
+        return nil;
       }
     }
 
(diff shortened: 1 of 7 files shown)


Sources/ATTNCustomEvent.m
```diff
@@ -18,7 +18,6 @@ - (instancetype)initWithType:(NSString *)type properties:(NSDictionary<NSString
     NSString *invalidCharInType = [ATTNCustomEvent findInvalidCharacterInType:type];
     if (invalidCharInType != nil) {
       NSLog(@"Invalid character '%@' in CustomEvent type '%@'", invalidCharInType, type);
-      [NSException raise:@"Bad CustomEvent Type" format:@"The CustomEvent type '%@' had an invalid character '%@'.", type, invalidCharInType];
```
oliviakim321, line 21: I think if we do encounter an exception like this here and below, we'd want to not create the event & return null or something

Sources/ATTNParameterValidation.m
```diff
@@ -52,7 +52,8 @@ + (void)verify1DStringDictionaryOrNil:(nullable NSDictionary *)inputValue inputN
     return;
 
   if (![inputValue isKindOfClass:[NSDictionary class]]) {
-    [NSException raise:@"Bad Identifier" format:@"%@ should be of form NSDictionary<NSString *, NSString *> *", inputName];
+    NSLog(@"Bad identifier: %@ should be of form NSDictionary<NSString *, NSString *> *", inputName);
+    return;
```
oliviakim321, line 56: I think we'll probably want to re-write these methods to return a bool or something to tell us whether the validation was successful & handle whether or not we create events etc at the call site? Since now these will just fail silently and cause errors downstream as opposed to stopping the flow

mrhota: I thought we wanted to just fire-and-forget--even if they're bad events-and let the backend just swallow them

oliviakim321 (approved):

PR #67, opened 2023-08-16
Update regex to better parse attn domains

rsmith-attentive:
Update regex to account for dtags that are compiled with the attn domain set to a variable, instead of directly on the window.

Dtags come from compiled JS, so when we update the compilation, the naming of the attn domain variable might change. We need to insulate our matching against this.

will extract domain successfully from:
- legacy dtag `<sampleDtagContent>,window.__attentive_domain='test-company-ca.attn.tv',`
- manually edited dtag `a='testcompany.attn.tv',<sampleDtagContent>,window.__attentive_domain='testcompany.attn.tv',`
- modern dtag `<sampleDtagContent>,a='test-company.attn.tv',`

### DIFF
--- a/Sources/ATTNAPI.m
+++ b/Sources/ATTNAPI.m
@@ -510,8 +510,6 @@ static NSString* const EVENT_TYPE_CUSTOM_EVENT = @"ce";
     return nil;
   }
 
-  NSLog(@"Found attn domain is: %@", [tag substringWithRange:domainRange]);
-
   return [tag substringWithRange:domainRange];
 }
 
--- a/Sources/ATTNAPI.m
+++ b/Sources/ATTNAPI.m
@@ -483,7 +483,7 @@ static NSString* const EVENT_TYPE_CUSTOM_EVENT = @"ce";
 
 + (NSString* _Nullable)extractDomainFromTag:(NSString*)tag {
   NSError* error = nil;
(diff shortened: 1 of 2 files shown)


Sources/ATTNAPI.m
```diff
@@ -483,7 +483,7 @@ - (NSMutableDictionary*)buildBaseMetadata:(ATTNUserIdentity*)userIdentity {
 
 + (NSString* _Nullable)extractDomainFromTag:(NSString*)tag {
   NSError* error = nil;
-  NSRegularExpression* regex = [NSRegularExpression regularExpressionWithPattern:@"window.__attentive_domain='(.*?).attn.tv'" options:0 error:&error];
+  NSRegularExpression* regex = [NSRegularExpression regularExpressionWithPattern:@"='([a-z0-9-]+)[.]attn[.]tv'" options:0 error:&error];
```
mrhota, line 486: will this work with dev companies that have, for example, `siren.dev.attn.tv`?

wyattjoe40 (approved):

Sources/ATTNAPI.m
rsmith-attentive: It won't - that was a decision made early in the process afaik to avoid unnecessary code complexity on a public repo. We have a prod test company for that reason, AttentiveGames

PR #72, opened 2024-01-16
Implement auto-close creatives on nav away

rsmith-attentive:
- Auto closes creative on nav away
- Updates example app to have tab nav for testing the above behavior
- Adds scroll to example app
- adds test for new behavior
- More logging to help debugging efforts


Sources/ATTNSDK.m
```diff
@@ -100,9 +101,10 @@ - (void)trigger:(UIView *)theView handler:(ATTNCreativeTriggerCompletionHandler)
 
   [[wkWebViewConfiguration userContentController] addScriptMessageHandler:self name:@"log"];
 
-  NSString *userScriptWithEventListener = @"window.addEventListener('message', (event) => {if (event.data && event.data.__attentive && event.data.__attentive.action === 'CLOSE') {window.webkit.messageHandlers.log.postMessage(event.data.__attentive.action);}}, false);";
+  NSString *userScriptWithEventListener = @"window.addEventListener('message', (event) => {if (event.data && event.data.__attentive) {window.webkit.messageHandlers.log.postMessage(event.data.__attentive.action);}}, false);window.addEventListener('visibilitychange', (event) => {window.webkit.messageHandlers.log.postMessage(`Document Hidden: ${document.hidden}`);}, false);";
```
TWaltze, line 104: I suggest moving the message `Document Hidden: ${document.hidden}` to a variable/function that we can template. The message as it stands is a little too "rich" in my opinion (various capitalizations, punctuation, spacing, etc) to be manually matched against down below. Maybe generate the string using a function like `generateVisibilityEvent(visibility: boolean): string`

ExampleSwift/Podfile
```diff
@@ -5,6 +5,6 @@ target 'ExampleSwift - Pod' do
   use_frameworks!
 
   # Pods for Example
-  pod 'attentive-ios-sdk'
+  pod 'attentive-ios-sdk', '0.4.3-beta.1'
```
TWaltze, line 8: Are we always going to want to define the exact version for the example app now? So in future version changes we'll want to bump this as well?

Sources/ATTNSDK.m
rsmith-attentive: good point, will update

ExampleSwift/Podfile
rsmith-attentive: I don't have a strong conviction either way. This change was to test out the beta functionality of cocoapods, but I can revert now that we know it works as expected.

TWaltze (dismissed):

TWaltze (approved):

PR #83, opened 2024-05-20
[APP-2867] Fix project structure to push to cocoapods and SPM from one codebase

vespinola:


rsmith-attentive (approved):
!!! this is awesome!

jasonaden:
Chatted with Vald on this. Looks great!

My only request is to split out changing the version from meaningful code commits. Let's remove the version change from here and make that in a release commit.

vespinola:
> Chatted with Vald on this. Looks great!
> 
> My only request is to split out changing the version from meaningful code commits. Let's remove the version change from here and make that in a release commit.

Sure @jasonaden. I'll revert the bump version from here and then raise a new PR for that.

rsmith-attentive (approved):

PR #95, opened 2024-06-14
[APP-2804] Mobile App Domain Switching - iOS

vespinola:
## PR Type
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [ ] Tests
- [ ] Other

[Jira Link](https://attentivemobile.atlassian.net/browse/APP-2804)

## What's new?

Added a new way to update the SDK domain.

```swift
let sdk = ATTNSDK(domain: "oldDomain")
sdk.update(domain: "newDomain")
```

## Testing

* Added unit test to cover new public method
* Added one UI Automation Test Case

## Screenshots

https://github.com/attentive-mobile/attentive-ios-sdk/assets/11541905/0242372e-a4f7-4372-9ffe-619694279245

### DIFF
--- a/Sources/Public/SDK/ATTNSDK.swift
+++ b/Sources/Public/SDK/ATTNSDK.swift
@@ -279,8 +279,14 @@ extension ATTNSDK {
     self.urlBuilder = urlBuilder
   }
 
-  convenience init(api: ATTNAPIProtocol) {
+  convenience init(api: ATTNAPIProtocol, urlBuilder: ATTNCreativeUrlProviding? = nil) {
     self.init(domain: api.domain)
     self.api = api
+    guard let urlBuilder = urlBuilder else { return }
+    self.urlBuilder = urlBuilder
+  }
+
+  func getDomain() -> String {
+    domain
   }
 }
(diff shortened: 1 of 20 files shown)


Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -147,6 +147,14 @@ public final class ATTNSDK: NSObject {
   public func clearUser() {
     userIdentity.clearUser()
   }
+
+  @objc(updateDomain:)
+  public func update(domain: String) {
```
vespinola, line 152: Steps to update domain:
- The new domain **should not** be the same as the previous one
- Update SDK domain property
- Update API domain property
  - cachedGeoAdjustedDomain reset to nil to request the geo adjusted service again
- Trigger identity event with cached data

TWaltze: In this case "should not" means it'll just noop, right? Not that we'll see some sort of error.

vespinola: That's correct. The user will not receive errors. The execution will only stop if the new domain is the same as the one already configured.

TWaltze (approved):

PR #120, opened 2025-02-21
Fix Github Actions CI

masterginger:
* Use macos-14 runner
* Use actions/checkout@v4
* Use actions/upload-artifacts@v4
* Remove the customized DEVELOPER_DIR
* Use iOS 17.4 + iPhone 15 Pro as the run target


adelag-attentive (approved):
Thank you so much!

PR #177, opened 2026-02-03
Updates `Inbox` to be a backend provided model and renames class

usharif-attentive:
## PR Type
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## What's new?
* Updates models to include an `Inbox` model that contains messages and unread count
* Updates `inbox` to be `inboxManager`

## Testing
Build and run locally

## Screenshots
n/a


Sources/Inbox/InboxManager.swift
```diff
@@ -154,5 +154,6 @@ fileprivate extension Inbox {
                 actionURLString: "https://example.com/track/12345"
             )
         ]
+        return Inbox(messages: messages, unreadCount: messages.count)
```
adelag-attentive: let's use the actual unread count (can put in a hard coded value there to mock it) so we can simulate the fact that we should be getting the unread count from server response, not computing it locally. this will help us stay organized once we have the live endpoint

adelag-attentive (approved):

PR #231, opened 2026-04-03
MSDK-298: Improve SwiftLint config for code clarity and enforce linting

adelag-attentive:
## PR Type
- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [x] CI related changes
- [ ] Tests
- [ ] Other

[MSDK-298](https://attentivemobile.atlassian.net/browse/MSDK-298)

## What's new?

Overhauls the SwiftLint configuration to improve code clarity for AI-assisted development, and tightens lint enforcement so violations can't silently pass through CI.

### SwiftLint config (`.swiftlint.yml`)
- **16 new opt-in rules** focused on clarity: `explicit_top_level_acl`, `missing_docs`, `force_unwrapping`, `closure_body_length`, `discouraged_optional_collection`, `modifier_order`, `sorted_imports`, and more
- **Re-enabled size limits** (previously fully disabled): `file_length`, `line_length`, `function_body_length`, `type_body_length` — with calibrated thresholds set above current worst offenders
- See inline comments in `.swiftlint.yml` for rationale on each rule

### Lint enforcement
- **Fastlane**: `ignore_exit_status` changed to `false`, `raise_if_swiftlint_error` changed to `true` — CI now fails on error-level violations
- **Xcode build phase**: Missing SwiftLint now produces an error + `exit 1` instead of a silent warning

### Auto-fixed violations
- Sorted imports, modifier ordering, colon spacing, unused closure params, superfluous else, statement position, trailing whitespace

### Source fix
- Broke up a 410-character inline JavaScript string in `ATTNWebViewHandling.swift` into a readable multiline heredoc

### Current state
- **0 errors**, 105 warnings (all legitimate clarity improvements to address incrementally in Phase 3)
- All unit tests pass

## Testing
- `swiftlint lint` → 105 warnings, 0 errors
- `xcodebuild test` → all tests pass
- Verified fastlane lint lane behavior with new settings

[MSDK-298]: https://attentivemobile.atlassian.net/browse/MSDK-298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,7 +133,7 @@ commands:
     steps:
       - run:
           name: Install SwiftLint
-          command: brew install swiftlint
+          command: command -v swiftlint || brew install swiftlint
 
   setup-fastlane:
     description: "Install Fastlane via Bundler"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -288,7 +288,6 @@ jobs:
       - attach_workspace:
           at: .
 
-      - install-swiftlint
       - setup-fastlane
 
       - run:
@@ -305,8 +304,6 @@ jobs:
     steps:
       - attach_workspace:
           at: .
-
-      - install-swiftlint
 
       - aws-cli/setup:
           region: AWS_REGION
@@ -396,7 +393,6 @@ jobs:
       - attach_workspace:
           at: .
 
-      - install-swiftlint
       - setup-fastlane
 
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,6 +128,13 @@ commands:
               ]
             }
 
+  install-swiftlint:
+    description: "Install SwiftLint (required by Xcode build phase)"
+    steps:
+      - run:
+          name: Install SwiftLint
+          command: brew install swiftlint
+
   setup-fastlane:
     description: "Install Fastlane via Bundler"
     steps:
@@ -215,6 +222,7 @@ jobs:
       - attach_workspace:
           at: .
 
+      - install-swiftlint
       - setup-fastlane
 
       - run:
@@ -262,6 +270,7 @@ jobs:
       - attach_workspace:
           at: .
 
+      - install-swiftlint
       - setup-fastlane
 
       - run:
@@ -279,6 +288,7 @@ jobs:
       - attach_workspace:
           at: .
 
+      - install-swiftlint
       - setup-fastlane
 
       - run:
@@ -295,6 +305,8 @@ jobs:
     steps:
       - attach_workspace:
           at: .
+
+      - install-swiftlint
 
       - aws-cli/setup:
           region: AWS_REGION
@@ -326,6 +338,7 @@ jobs:
       - attach_workspace:
           at: .
 
+      - install-swiftlint
       - setup-fastlane
 
       - run:
@@ -383,6 +396,7 @@ jobs:
       - attach_workspace:
           at: .
 
+      - install-swiftlint
       - setup-fastlane
 
       - run:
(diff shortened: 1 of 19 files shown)


fastlane/Fastfile
```diff
@@ -221,7 +221,7 @@ platform :ios do
     swiftlint(
       mode: :lint,
       output_file: "#{FASTLANE_REPORTS_DIR}/swiftlint.xml",
-      raise_if_swiftlint_error: false,
+      raise_if_swiftlint_error: true,
```
usharif-attentive, line 224: `ignore_exit_status: true` is still set here, which tells the Fastlane SwiftLint action to ignore SwiftLint's exit code entirely. With this flag still `true`, `raise_if_swiftlint_error` only catches SwiftLint *crashes*, not lint violations that produce a non-zero exit. If the goal is to fail CI on error-level violations, `ignore_exit_status` should also be set to `false`.

.circleci/config.yml
```diff
@@ -128,6 +128,13 @@ commands:
               ]
             }
 
+  install-swiftlint:
+    description: "Install SwiftLint (required by Xcode build phase)"
+    steps:
```
usharif-attentive, line 133: `install-swiftlint` is added to 6 jobs, including ones that don't run SwiftLint directly (e.g., deploy-testflight, release). `command -v swiftlint || brew install swiftlint` will install on every run since CI images likely don't have it cached, adding ~30-60s overhead per job.

A few suggestions:
- Only add `install-swiftlint` to jobs that actually invoke SwiftLint (the lint job and build jobs with the Xcode build phase)
- Pin a specific version for reproducibility (e.g., download the binary directly or use `brew install swiftlint@0.x.x`)
- Consider caching the Homebrew install across runs

Sources/ATTNWebViewHandling.swift
```diff
@@ -139,7 +139,19 @@ class ATTNWebViewHandler: NSObject, ATTNWebViewHandling {
                 let configuration = WKWebViewConfiguration()
                 configuration.userContentController.add(self, name: Constants.scriptMessageHandlerName)
 
-                let userScriptWithEventListener = String(format: "window.addEventListener('message', (event) => {if (event.data && event.data.__attentive) {window.webkit.messageHandlers.log.postMessage(event.data.__attentive.action);}}, false);window.addEventListener('visibilitychange', (event) => {window.webkit.messageHandlers.log.postMessage(`%@ ${document.hidden}`);}, false);", Constants.visibilityEvent)
+                let jsEventListeners = """
```
usharif-attentive: The indented multiline string literal will include the leading whitespace in the actual JS string at runtime. The original single-line string had no extra whitespace. Since this is injected as a `WKUserScript` it likely still works, but it's a subtle behavioral change — the JS now contains extra spaces/newlines that weren't there before. Consider trimming whitespace or removing the indentation to keep the runtime string identical.

fastlane/Fastfile
adelag-attentive, line 224: Re: `ignore_exit_status` (comment on Fastfile line 224) — Good callout. We tried `ignore_exit_status: false` but it fails on *any* violation (including warnings), not just error-level ones. The fastlane swiftlint action has no severity-based gating. With 105 intentional warnings, it always fails.

The enforcement model is: this lint lane generates JUnit reports for visibility, and the **Xcode build phase** is the actual enforcement point (it correctly fails only on error-level violations via SwiftLint's exit code). So `ignore_exit_status: true` is intentional here.

Also addressed the other two comments:
- **JS string whitespace** — switched from multiline heredoc to string concatenation, preserving identical runtime output
- **install-swiftlint scope** — removed from 3 jobs that don't build the SDK from source (build-bonni-with-binary, build-bonni-release, validate-spm). Kept on build-framework, unit-tests, and build-examples.

usharif-attentive (approved):

PR #232, opened 2026-04-03
MSDK-271: Add swift-snapshot-testing dependency and base test infrastructure

usharif-attentive:
## Summary

- Added [swift-snapshot-testing](https://github.com/pointfreeco/swift-snapshot-testing) as a test-only SPM dependency in `Package.swift` and as a package in the Xcode project
- Created `ATTNSnapshotTestCase` base class with shared configuration for device sizes (iPhone SE, iPhone 16, iPhone 16 Pro Max, iPad mini) and `__Snapshots__` reference image directory convention
- CI-safe: uses `record: .never` when `CI` env var is set, so missing snapshots fail instead of silently recording
- Excluded `.build` from SwiftLint to prevent lint errors on dependency sources

## Jira

[MSDK-271](https://attentivemobile.atlassian.net/browse/MSDK-271)

## Test plan

- [x] Existing unit tests pass (`bundle exec fastlane ios unit_test`)
- [ ] Verify Xcode GUI build succeeds with snapshot dependency resolved
- [ ] Subclass `ATTNSnapshotTestCase` and confirm snapshot recording works locally

[MSDK-271]: https://attentivemobile.atlassian.net/browse/MSDK-271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -143,3 +143,4 @@ excluded:
   - vendor
   - Vendor
   - .build
+  - fastlane/build
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -142,5 +142,4 @@ excluded:
   - Pods
   - vendor
   - Vendor
-  - Tests
   - .build
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -142,3 +142,5 @@ excluded:
   - Pods
   - vendor
   - Vendor
+  - Tests
+  - .build
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,0 +1,41 @@
+{
+  "pins" : [
+    {
+      "identity" : "swift-custom-dump",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/pointfreeco/swift-custom-dump",
+      "state" : {
+        "revision" : "06c57924455064182d6b217f06ebc05d00cb2990",
+        "version" : "1.5.0"
+      }
+    },
+    {
+      "identity" : "swift-snapshot-testing",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/pointfreeco/swift-snapshot-testing",
+      "state" : {
+        "revision" : "05b6f05b55ff9e2dcdcc21ac78c6ba81ea624791",
+        "version" : "1.19.1"
+      }
+    },
+    {
+      "identity" : "swift-syntax",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/swiftlang/swift-syntax",
+      "state" : {
+        "revision" : "4799286537280063c85a32f09884cfbca301b1a1",
+        "version" : "602.0.0"
+      }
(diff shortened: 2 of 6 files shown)


.swiftlint.yml
```diff
@@ -19,4 +19,5 @@ nesting:
     error: 10
 excluded:
   - Example
-  - Tests
\ No newline at end of file
+  - Tests
+  - .build
```
adelag-attentive: Nit: the file still has no trailing newline (`\ No newline at end of file`). Most editors and linters expect one — worth adding to keep diffs clean.

Package.resolved
```diff
@@ -0,0 +1,41 @@
...
+        "version" : "1.5.0"
+      }
+    },
+    {
+      "identity" : "swift-snapshot-testing",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/pointfreeco/swift-snapshot-testing",
+      "state" : {
+        "revision" : "05b6f05b55ff9e2dcdcc21ac78c6ba81ea624791",
+        "version" : "1.19.1"
```
adelag-attentive: The root `Package.resolved` pins `swift-snapshot-testing` at **1.19.1** and `swift-syntax` at **602.0.0**, but the Xcode workspace copy (`attentive-ios-sdk.xcodeproj/…/Package.resolved`) pins **1.19.2** and **603.0.0**. This mismatch means SPM-only consumers and Xcode consumers resolve different dependency graphs. Worth reconciling so both lockfiles agree on the same versions.

Tests/TestCases/ATTNSnapshotTestCase.swift
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    private var recordMode: SnapshotTestingConfiguration.Record {
+        guard !Self.isCI else { return .never }
+        return isRecordingSnapshots ? .all : .missing
+    }
+
+    /// Asserts a snapshot of the given view controller for each provided device size.
+    /// Reference images are stored alongside the test file in a `__Snapshots__` directory.
+    func assertSnapshot(
```
adelag-attentive, line 77: When `isRecordingSnapshots` is `false` (the default) and we're not on CI, `recordMode` falls through to `.missing`, which silently records any snapshot that doesn't have a reference image yet — the test passes without anyone visually verifying the output.

In practice this means a developer can add a new snapshot test, run it locally, and get a green checkmark before ever looking at the generated image. Consider defaulting to `.never` instead (so missing references fail loudly and the dev must explicitly opt in to recording), or at minimum documenting this behavior so reviewers know to check for newly recorded images in the PR diff.

adelag-attentive (commented):
A few observations from reviewing the snapshot testing infrastructure:

adelag-attentive (approved):

PR #233, opened 2026-04-07
MSDK-264: Detach push token from user on clearUser

adelag-attentive:
## PR Type
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [x] Tests
- [ ] Other

[MSDK-264](https://attentivemobile.atlassian.net/browse/MSDK-264)

## What's new?

`clearUser()` now detaches the push token from the logged-out user by calling the `updateUser` API with empty email/phone identifiers. This ensures the previous user no longer receives push notifications on this device after logout.

**Implementation details:**
- Extracted `clearUserIdentifiers()` as a private helper to avoid a redundant API call when `updateUser()` internally clears identifiers
- If no push token is available, `clearUser()` gracefully skips the server-side detach and still clears identifiers locally
- Added `operationContext` parameter to `updateUser()` so log messages display the correct public API name (`clearUser` vs `updateUser`) since both share the same `/user-update` endpoint
- Fixed noisy identifier validation — `validate()` now only checks keys actually provided by the caller instead of iterating all possible identifier keys
- Updated README to document the push token detach behavior for users who have push notifications enabled

## Testing

- `testClearUser_withPushToken_callsUpdateUserWithNilEmailAndPhone` — verifies API is called with nil email/phone to detach token
- `testClearUser_withoutPushToken_doesNotCallUpdateUser` — verifies graceful no-op without a token
- `testClearUser_resetsIdentifiers` — verifies identifiers are still cleared
- `testUpdateUser_callsUpdateUserExactlyOnce` — verifies no double API call from the refactored code path

Screenshots from attentive UI: see [slack thread](https://attentivemobile.slack.com/archives/C087XAP637X/p1776356914678239)
All 171 tests pass.

[MSDK-264]: https://attentivemobile.atlassian.net/browse/MSDK-264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Bonni/AttentiveExample/SettingsViewController.swift
+++ b/Bonni/AttentiveExample/SettingsViewController.swift
@@ -388,7 +388,6 @@ class SettingsViewController: UIViewController {
             self.currentEmailLabel.text = "Current email: \(self.currentEmail ?? "")"
             self.currentPhoneLabel.text = "Current phone: \(self.currentPhone ?? "")"
 
-            self.getAttentiveSdk().clearUser()
             self.getAttentiveSdk().updateUser(
                 email: self.currentEmail,
                 phone: self.currentPhone
--- a/Bonni/AttentiveExample/SettingsViewController.swift
+++ b/Bonni/AttentiveExample/SettingsViewController.swift
@@ -65,6 +65,14 @@ class SettingsViewController: UIViewController {
         return button
     }()
 
+    private let logOutButton: UIButton = {
+        let button = UIButton(type: .system)
+        button.setTitle("Log Out", for: .normal)
+        button.setTitleColor(.systemRed, for: .normal)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        return button
+    }()
+
     // MARK: - Test Events Section
 
     private let showCreativeButton: UIButton = {
@@ -250,6 +258,7 @@ class SettingsViewController: UIViewController {
         stackView.addArrangedSubview(accountInfoLabel)
         stackView.addArrangedSubview(switchAccountButton)
         stackView.addArrangedSubview(switchDomainButton)
+        stackView.addArrangedSubview(logOutButton)
 
         let divider = UIView()
         divider.backgroundColor = .lightGray
@@ -272,6 +281,7 @@ class SettingsViewController: UIViewController {
             let allButtons: [UIButton] = [
                 switchAccountButton,
                 switchDomainButton,
+                logOutButton,
                 showCreativeButton,
                 showPushPermissionButton,
                 sendLocalPushNotification,
@@ -320,6 +330,7 @@ class SettingsViewController: UIViewController {
     private func setupActions() {
         switchAccountButton.addTarget(self, action: #selector(switchAccountTapped), for: .touchUpInside)
         switchDomainButton.addTarget(self, action: #selector(switchDomainTapped), for: .touchUpInside)
+        logOutButton.addTarget(self, action: #selector(logOutTapped), for: .touchUpInside)
         showCreativeButton.addTarget(self, action: #selector(showCreativeTapped), for: .touchUpInside)
         showPushPermissionButton.addTarget(self, action: #selector(showPushPermissionTapped), for: .touchUpInside)
         sendLocalPushNotification.addTarget(self, action: #selector(sendLocalPushNotificationTapped), for: .touchUpInside)
@@ -493,8 +504,19 @@ class SettingsViewController: UIViewController {
         showToast(with: success ? "Deep link handled: bonni://cart" : "Deep link failed")
     }
 
+    @objc private func logOutTapped() {
+        getAttentiveSdk().clearUser()
+        currentEmail = nil
+        currentPhone = nil
+        currentEmailLabel.text = "Current email:"
+        currentPhoneLabel.text = "Current phone:"
+        accountInfoLabel.text = "Login Info: Guest"
+        showToast(with: "Logged out — push token detached")
+    }
+
     @objc private func clearUserTapped() {
-        // TODO: Clear the user
+        getAttentiveSdk().clearUser()
+        showToast(with: "User cleared")
     }
 
     @objc private func clearCookiesTapped() {
--- a/Sources/Public/SDK/ATTNSDK.swift
+++ b/Sources/Public/SDK/ATTNSDK.swift
@@ -167,6 +167,27 @@ public final class ATTNSDK: NSObject {
 
     @objc(clearUser)
     public func clearUser() {
+        clearUserIdentifiers()
+
+        // Detach push token from the logged-out user by calling updateUser with empty identifiers
+        let pushToken = currentPushToken
+        guard !pushToken.isEmpty else {
+            Loggers.event.debug("clearUser: skipping push token detach — no push token available")
+            return
+        }
+
+        Loggers.event.debug("clearUser: detaching push token from previous user - Visitor ID: \(self.userIdentity.visitorId, privacy: .public)")
+        api.updateUser(
+            pushToken: pushToken,
+            userIdentity: userIdentity,
+            email: nil,
+            phone: nil,
+            callback: nil
+        )
+    }
+
+    /// Clears user identifiers and generates a new visitor ID without making any network calls.
+    private func clearUserIdentifiers() {
         let oldVisitorId = userIdentity.visitorId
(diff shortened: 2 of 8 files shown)


usharif-attentive (commented):
(see inline comments)

Bonni/AttentiveExample/SettingsViewController.swift
```diff
@@ -493,8 +504,19 @@ class SettingsViewController: UIViewController {
         showToast(with: success ? "Deep link handled: bonni://cart" : "Deep link failed")
     }
 
+    @objc private func logOutTapped() {
+        getAttentiveSdk().clearUser()
```
usharif-attentive, line 507: This new `logOutTapped` method looks good, but note that the existing `switchAccountTapped` method (~line 391, not in this diff) still calls `clearUser()` explicitly before `updateUser()`. With this PR, that `clearUser()` now fires a network request to detach the push token, and then `updateUser()` immediately re-attaches it — that's a wasted round-trip and could briefly put the server in a confusing state (detach → immediate re-attach).

The README was already updated to remove `clearUser()` before `updateUser()`, but `switchAccountTapped` in the demo app wasn't updated to match. Suggest removing the `clearUser()` call there since `updateUser()` already calls `clearUserIdentifiers()` internally.

Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -165,8 +165,52 @@ public final class ATTNSDK: NSObject {
...
+    ///   detach is skipped — identifiers are still cleared locally.
+    ///
+    /// Internal implementation detail (for maintainers / AI assistants):
+    /// Under the hood this calls the same `/user-update` endpoint as `updateUser`, but with
+    /// empty email and phone. The `operationContext: "clearUser"` parameter ensures all logs
+    /// are labelled "clearUser" so SDK consumers never see "updateUser" in their console when
+    /// they only called `clearUser()`.
     @objc(clearUser)
     public func clearUser() {
+        clearUserIdentifiers()
```
usharif-attentive, line 189: Identifiers are cleared and a new visitor ID is generated *before* the API call. The detach request is then sent with the **new** anonymous visitor ID, not the old one. If the server needs the old visitor ID to identify which user to detach the token from, this could be a bug.

Can you confirm the server-side contract here? (Same pattern exists in `updateUser()` so it may be intentional — just want to verify.)

Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -165,8 +165,52 @@ public final class ATTNSDK: NSObject {
...
+
+        let pushToken = currentPushToken
```
usharif-attentive, line 191: `clearUser()` uses `currentPushToken` directly with no trimming or fallback:
```swift
let pushToken = currentPushToken
guard !pushToken.isEmpty else { ... }
```

But `updateUser()` trims whitespace and falls back to UserDefaults:
```swift
let trimmedPushToken = currentPushToken.trimmingCharacters(in: .whitespacesAndNewlines)
let pushToken = !trimmedPushToken.isEmpty
    ? trimmedPushToken
    : (UserDefaults.standard.string(forKey: "attentiveDeviceToken") ?? "")
```

If the in-memory token has whitespace or is missing but present in UserDefaults, `clearUser` would skip the detach while `updateUser` would succeed. These should use the same retrieval logic — consider extracting a shared helper.

Tests/Doubles/Spies/ATTNAPISpy.swift
```diff
@@ -128,16 +129,21 @@ final class ATTNAPISpy: ATTNAPIProtocol {
...
         email: String?,
         phone: String?,
+        operationContext: String,
         callback: ATTNAPICallback?
     ) {
         updateUserWasCalled = true
+        updateUserCallCount += 1
         lastUpdateUserEmail = email
         lastUpdateUserPhone = phone
+        lastOperationContext = operationContext
```
usharif-attentive, line 148: The spy captures `lastUpdateUserEmail`, `lastUpdateUserPhone`, and `lastOperationContext`, but doesn't capture the `pushToken` parameter. This means tests can't assert that `clearUser` passes the correct push token to the API.

Suggest adding:
```swift
lastPushToken = pushToken
```

Tests/TestCases/ATTNSDKTests.swift
```diff
@@ -368,6 +368,50 @@ final class ATTNSDKTests: XCTestCase {
...
+        let deviceToken = Data([0x01, 0x02, 0x03])
+        sut.registerDeviceToken(deviceToken, authorizationStatus: .authorized)
+
+        XCTAssertFalse(apiSpy.updateUserWasCalled)
+
+        sut.clearUser()
+
+        XCTAssertTrue(apiSpy.updateUserWasCalled, "clearUser should call updateUser to detach push token")
+        XCTAssertNil(apiSpy.lastUpdateUserEmail, "clearUser should pass nil email to detach push token")
+        XCTAssertNil(apiSpy.lastUpdateUserPhone, "clearUser should pass nil phone to detach push token")
```
usharif-attentive, line 383: The spy captures `lastOperationContext` but no test asserts its value. Since the whole point of the parameter is to label logs correctly per the calling context, consider adding:
```swift
XCTAssertEqual(apiSpy.lastOperationContext, "clearUser")
```
here, and similarly `"updateUser"` in `testUpdateUser_callsUpdateUserExactlyOnce`.

usharif-attentive (commented):
Thanks for the PR — the push token detach on logout is a solid feature. Left a few inline comments.

Bonni/AttentiveExample/SettingsViewController.swift
adelag-attentive: AI is confused. i'll update docs

Sources/Public/SDK/ATTNSDK.swift
adelag-attentive: another reason to get the plugin setup soon - claude is confused without back end context

Sources/Public/SDK/ATTNSDK.swift
adelag-attentive: Pushing back on this one. `clearUser()` uses `currentPushToken` which goes through `PushTokenStore.token` — and that getter already falls back to UserDefaults:

```swift
var token: String {
    get {
        queue.sync {
            if let token = inMemory, !token.isEmpty { return token }
            let persisted = UserDefaults.standard.string(forKey: key) ?? ""
            inMemory = persisted
            return persisted
        }
    }
}
```

So both code paths already check UserDefaults via `PushTokenStore`. The extra `UserDefaults.standard.string(forKey: "attentiveDeviceToken")` fallback in `updateUser()` is actually redundant pre-existing code that never triggers. And trimming a hex token string (`"%02.2hhx"` joined) is a no-op since hex digits can never contain whitespace.

No change needed here — both methods use the same underlying `PushTokenStore` which handles the fallback.

Tests/Doubles/Spies/ATTNAPISpy.swift
adelag-attentive: Done — added `lastUpdateUserPushToken` to the spy.

Tests/TestCases/ATTNSDKTests.swift
adelag-attentive: Done — added `operationContext` assertions to both tests:
- `testClearUser_withPushToken_callsUpdateUserWithNilEmailAndPhone` now asserts `"clearUser"`
- `testUpdateUser_callsUpdateUserExactlyOnce` now asserts `"updateUser"`

Also added `lastUpdateUserPushToken` assertion in the clearUser test.

Bonni/AttentiveExample/SettingsViewController.swift
adelag-attentive: Good catch on `switchAccountTapped` — removed the redundant `clearUser()` call since `updateUser()` already calls `clearUserIdentifiers()` internally.

usharif-attentive: Verified — the redundant `clearUser()` call is gone in `0e4e022b`. Thanks!

usharif-attentive (approved):

PR #236, opened 2026-04-10
MSDK-319: Skip geo-domain adjustment for mobile SDK

adelag-attentive:
## PR Type
- [x] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Tests
- [ ] Other

[Jira Link](https://attentivemobile.atlassian.net/browse/MSDK-319)

## What's new?

The mobile SDK now uses the raw domain directly for all API requests instead of fetching the geo-adjusted domain from `cdn.attn.tv/{domain}/dtag.js`. This fixes a bug where regionalized domains (e.g., `youngla-pt`) would get double-suffixed by the CDN router (e.g., `youngla-pt-pt`), causing a 404/204 and silently dropping all events, push tokens, and subscription requests.

### Changes
- **`ATTNAPI.swift`** — Removed all geo-domain adjustment infrastructure: `getGeoAdjustedDomain()`, `cachedGeoAdjustedDomain`, `extractDomainFromTag()`, `RequestConstants` (dtag URL + regex pattern), and `enableGeoAdjustmentFallback()`. All callers (`send`, `sendPushToken`, `sendOptIn/OutMarketingSubscription`, `sendNewEvent`) now use the raw `domain` directly instead of going through the async geo-adjustment callback.
- **`ATTNAPIProtocol.swift`** — Removed `cachedGeoAdjustedDomain` from protocol
- **`ATTNError.swift`** — Kept `.geoDomainUnavailable` case but marked it `@available(*, deprecated)` to preserve public API compatibility for consumers with exhaustive switches
- **`ATTNAPITests.swift`** — Removed geo-specific tests and dtag assertions; simplified multiple-event test
- **`ATTNAPIITTests.swift`** — Replaced `testGeoAdjustedDomain` with direct `testDomain` usage
- **`ATTNTestEventUtils.swift`** — Renamed `geoAdjustedDomain` parameter to `domain`
- **`NSURLSessionMock.swift`** — Removed dtag CDN mock handling
- **`ATTNAPISpy.swift`** — Removed `cachedGeoAdjustedDomain` tracking

## Testing

- All 167 unit tests pass with 0 failures
- Verified all API methods use the raw domain directly

### DIFF
--- a/Sources/API/ATTNError.swift
+++ b/Sources/API/ATTNError.swift
@@ -10,7 +10,6 @@ import Foundation
 public enum ATTNError: Error {
     case sdkNotInitialized
     case missingContactInfo
-    case geoDomainUnavailable
     case badURL
     case invalidDomain
 }
@@ -22,8 +21,6 @@ extension ATTNError: LocalizedError {
             return "SDK not initialized"
         case .missingContactInfo:
             return "Provide email and/or phone"
-        case .geoDomainUnavailable:
-            return "Geo domain unavailable"
         case .badURL:
             return "Invalid URL"
         case .invalidDomain:
--- a/Sources/API/ATTNError.swift
+++ b/Sources/API/ATTNError.swift
@@ -10,6 +10,8 @@ import Foundation
 public enum ATTNError: Error {
     case sdkNotInitialized
     case missingContactInfo
+    @available(*, deprecated, message: "Geo-domain adjustment has been removed. This case is no longer used by the SDK.")
+    case geoDomainUnavailable
     case badURL
     case invalidDomain
 }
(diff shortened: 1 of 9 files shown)


adelag-attentive:
@codex review again

Sources/API/ATTNError.swift
```diff
@@ -10,7 +10,6 @@ import Foundation
 public enum ATTNError: Error {
     case sdkNotInitialized
     case missingContactInfo
-    case geoDomainUnavailable
```
usharif-attentive, line 13: **Breaking Public API Change**: `ATTNError` is a `public enum`, so removing `.geoDomainUnavailable` is a source-breaking change for any SDK consumer that has an exhaustive `switch` on `ATTNError` or explicitly references that case — their code will fail to compile after upgrading.

Consider either:
- Keeping the case but marking it as deprecated: `@available(*, deprecated, message: "No longer used")`
- Or treating this as a semver major version bump

This would let existing consumers upgrade without a compile error while signaling the case is no longer meaningful.

usharif-attentive (approved):

PR #242, opened 2026-04-17
MSDK-321: Adds `settings.json` for Claude

usharif-attentive:
## PR Type
- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [x] Other

https://attentivemobile.atlassian.net/browse/MSDK-321

## What's new?
Adds `settings.json` for Claude

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,13 @@
+{
+  "extraKnownMarketplaces": {
+    "attentive-marketplace": {
+      "source": {
+        "source": "github",
+        "repo": "attentive-mobile/claude-plugins"
+      }
+    }
+  },
+  "enabledPlugins": {
+    "mobile-sdk-internal@attentive-marketplace": true
+  }
+}


adelag-attentive:
**Nested `source.source` — is this the correct schema?** The structure has `"source": { "source": "github", "repo": "..." }`. The outer `source` is the key name and the inner `source` is the type discriminator. Can you confirm this is the expected format per the Claude Code marketplace config docs (vs. the outer key being something like `"config"`)?

**Plugin validation:** Is `mobile-sdk-internal@attentive-marketplace` already published in the `attentive-mobile/claude-plugins` repo? If not, enabling it here before the plugin exists would be a no-op (harmless but potentially confusing).

usharif-attentive:
> **Nested `source.source` — is this the correct schema?** The structure has `"source": { "source": "github", "repo": "..." }`. The outer `source` is the key name and the inner `source` is the type discriminator. Can you confirm this is the expected format per the Claude Code marketplace config docs (vs. the outer key being something like `"config"`)?
> 
> **Plugin validation:** Is `mobile-sdk-internal@attentive-marketplace` already published in the `attentive-mobile/claude-plugins` repo? If not, enabling it here before the plugin exists would be a no-op (harmless but potentially confusing).

On the schema: Confirmed correct — the nested source.source is the documented pattern. extraKnownMarketplaces.{name}.source is the wrapper object, and the inner source: "github" is the type discriminator (alongside repo). Docs: https://code.claude.com/docs/en/plugin-marketplaces.md

On the plugin validation: Correct, the plugin is not available yet, but as stated, it would be a harmless no-op to attempt to access it without it being published, so this is ok for now.

adelag-attentive (approved):

PR #244, opened 2026-04-20
Security fixes to CircleCI pipeline

bpeters-mcis:
## PR Type
- [X] CI related changes

## What's new?

These are minor changes to address some security findings on the CircleCI pipeline:

1. Instead of passing the `GITHUB_TOKEN` in the clone URL, it passes it as an additional header.  This prevents the token from possibly showing up in log files or other places it should not be visible
2. Moved the CircleCI AWS role ARN to an environmental variable.  While hard coding it was super low risk, it does present a tiny chance of providing intel to malicious actors, and the fix is simple, so we might as well do it

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -152,7 +152,7 @@ jobs:
       - run:
           name: Checkout code
           command: |
-            git -c "http.https://github.com/.extraheader=Authorization: basic $(printf 'x-access-token:%s' "${GITHUB_TOKEN}" | base64)" \
+            git -c "http.https://github.com/.extraheader=Authorization: basic $(printf 'x-access-token:%s' "${GITHUB_TOKEN}" | base64 -w 0)" \
               clone --filter=blob:none \
               "https://github.com/attentive-mobile/attentive-ios-sdk.git" .
             git checkout "${CIRCLE_SHA1}"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -152,8 +152,9 @@ jobs:
       - run:
           name: Checkout code
           command: |
-            git clone --filter=blob:none \
-              "https://x-access-token:${GITHUB_TOKEN}@github.com/attentive-mobile/attentive-ios-sdk.git" .
+            git -c "http.https://github.com/.extraheader=Authorization: basic $(printf 'x-access-token:%s' "${GITHUB_TOKEN}" | base64)" \
```
usharif-attentive: `base64` on CircleCI's default Linux images (GNU coreutils) wraps output at 76 chars by default. `x-access-token:<token>` for a classic PAT (~40 chars) encodes to ~76 chars — right at the boundary — and longer tokens (fine-grained PATs, GitHub App installation tokens like `ghs_...`) will wrap, producing an `Authorization` header with an embedded newline and failing auth.

```suggestion
            git -c "http.https://github.com/.extraheader=Authorization: basic $(printf 'x-access-token:%s' "${GITHUB_TOKEN}" | base64 -w 0)" \
```

usharif-attentive (changes requested):

usharif-attentive (dismissed):

usharif-attentive (approved):

PR #252, opened 2026-05-01
MSDK-337: Fix Bonni app v2 endpoint events

adelag-attentive:
## Summary

- **Fix duplicate events:** V2 buttons (Add to Cart, Place Order) in the Bonni app were firing both a `/mobile` (v2) and `/e` (v1) event. Moved legacy event sending out of the delegate callback so each button owns its own event path.
- **App-wide v1/v2 toggle:** Added `useV2Endpoint` property on `ATTNSDK` that automatically converts legacy `record(event:)` calls to v2 format and routes them through `/mobile`. Supports Purchase, AddToCart, ProductView, and CustomEvent; unsupported types (e.g. InfoEvent) fall through to `/e`.
- **ProductView on product tap:** Tapping a product cell in the grid now fires a ProductView event, routed by the toggle.
- **Settings toggle with persistence:** Added a toggle switch in Bonni's Settings screen to flip between v1/v2 at runtime, persisted via UserDefaults across app launches.

Jira: [MSDK-337](https://attentivemobile.atlassian.net/browse/MSDK-337)

## Test plan

- [x] Verified via Datadog logs that V2 buttons send only `/mobile` events (no duplicate `/e` events)
- [x] Verified via Datadog logs that with toggle ON, legacy `record(event:)` calls route through `/mobile`
- [x] Verified identity (`t=i`) events still go through `/e` as expected
- [ ] Toggle ON → tap product → verify ProductView goes to `/mobile`
- [ ] Toggle OFF → tap product → verify ProductView goes to `/e`
- [ ] Toggle ON → Place Order (legacy button) → verify Purchase goes to `/mobile`
- [ ] Kill and relaunch app → verify toggle state persists

[MSDK-337]: https://attentivemobile.atlassian.net/browse/MSDK-337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/API/ATTNEventTypes.swift
+++ b/Sources/API/ATTNEventTypes.swift
@@ -178,9 +178,11 @@ public struct ATTNPurchaseMetadata: Codable {
 /// Metadata for "MobileCustomEvent" events.
 public struct ATTNMobileCustomEventMetadata: Codable {
     public let eventType: String = "MobileCustomEvent"
+    public let type: String?
     public let customProperties: [String: String]?
 
-    public init(customProperties: [String: String]? = nil) {
+    public init(type: String? = nil, customProperties: [String: String]? = nil) {
+        self.type = type
         self.customProperties = customProperties
     }
 }
--- a/Sources/Helpers/Extension/ATTNSDK+Extension.swift
+++ b/Sources/Helpers/Extension/ATTNSDK+Extension.swift
@@ -9,15 +9,19 @@ import Foundation
 
 extension ATTNSDK {
     func send(event: ATTNEvent) {
-        if useV2Endpoint, let v2Event = convertLegacyEventToV2(event) {
-            recordV2EventData(v2Event)
+        if useV2Endpoint {
+            sendLegacyEventAsV2(event)
             return
         }
         api.send(event: event, userIdentity: userIdentity)
     }
 
-    private func convertLegacyEventToV2(_ event: ATTNEvent) -> ATTNEventData? {
+    private func sendLegacyEventAsV2(_ event: ATTNEvent) {
         if let purchase = event as? ATTNPurchaseEvent {
+            guard !purchase.items.isEmpty else {
+                Loggers.event.debug("No items found in the purchase event, skipping v2 send.")
+                return
+            }
             let products = purchase.items.map { item in
                 ATTNProduct(
                     productId: item.productId,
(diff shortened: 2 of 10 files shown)


tpfaff (dismissed):

Sources/API/ATTNEventTypes.swift
```diff
@@ -178,9 +178,11 @@ public struct ATTNPurchaseMetadata: Codable {
 /// Metadata for "MobileCustomEvent" events.
 public struct ATTNMobileCustomEventMetadata: Codable {
     public let eventType: String = "MobileCustomEvent"
+    public let type: String?
```
usharif-attentive, line 181: **🔴 Verify backend `/mobile` contract for the new `type` field**

`ATTNMobileCustomEventMetadata` now serializes both `eventType` and `type`. Worth cross-checking against the `mobile-api/` service definition before merge:

- If the backend requires `type` on `MobileCustomEvent`, every existing `sendCustomEvent(customProperties:)` call-site silently sends `null` the moment the toggle flips on.
- If `type` is additive, this is fine — but since it's public on a public `Codable` struct, it's a public-API surface change. Intentional?

Sources/Helpers/Extension/ATTNSDK+Extension.swift
```diff
@@ -9,6 +9,91 @@ import Foundation
...
 extension ATTNSDK {
     func send(event: ATTNEvent) {
+        if useV2Endpoint {
+            sendLegacyEventAsV2(event)
+            return
+        }
+        api.send(event: event, userIdentity: userIdentity)
+    }
+
+    private func sendLegacyEventAsV2(_ event: ATTNEvent) {
```
usharif-attentive, line 19: **🟠 Add unit tests for `sendLegacyEventAsV2`**

This function has real branching (per-event-type dispatch, empty-items guards, currency defaulting, orderTotal math, legacy fallthrough) but no new tests. CLAUDE.md calls out that new public API should have corresponding unit tests — and the fan-out logic here is exactly the kind of thing that silently regresses.

Suggested coverage:
- `useV2Endpoint = true` + each event type → verify `ATTNAPISpy.sendNewEvent` is called with expected metadata
- `useV2Endpoint = true` + unsupported type (e.g. `ATTNInfoEvent`) → verify legacy `api.send(...)` still fires
- Empty-items purchase/addToCart/productView → verify neither API method is called
- `useV2Endpoint = false` → verify legacy `api.send(...)` is called unchanged

Sources/Helpers/Extension/ATTNSDK+Extension.swift
```diff
@@ -9,6 +9,91 @@ import Foundation
...
+            let products = purchase.items.map { item in
+                ATTNProduct(
+                    productId: item.productId,
+                    variantId: item.productVariantId,
+                    name: item.name ?? "",
+                    imageUrl: item.productImage,
+                    categories: item.category.map { [$0] },
+                    price: item.price.price.stringValue,
+                    quantity: item.quantity
+                )
```
usharif-attentive: **🟢 Extract ATTNItem → ATTNProduct helper**

This same 8-line mapping is repeated in the AddToCart and ProductView branches below. Extracting a small helper removes ~20 lines of duplication and makes future `ATTNProduct` field additions a one-line change:

```swift
private func product(from item: ATTNItem) -> ATTNProduct {
    ATTNProduct(
        productId: item.productId,
        variantId: item.productVariantId,
        name: item.name ?? "",
        imageUrl: item.productImage,
        categories: item.category.map { [$0] },
        price: item.price.price.stringValue,
        quantity: item.quantity
    )
}
```

Sources/Helpers/Extension/ATTNSDK+Extension.swift
```diff
@@ -9,6 +9,91 @@ import Foundation
...
+            }
+            let cart = ATTNCartPayload(from: purchase.cart)
+            let currency = purchase.items.first?.price.currency ?? "USD"
```
usharif-attentive: **🟡 Unify currency fallback policy across event types**

This defaults to `"USD"`, but the AddToCart and ProductView branches below use `item.price.currency` directly with no fallback. Pick one policy and apply everywhere:

- If items always have a currency, the `?? "USD"` here is dead code.
- If they can be empty, AddToCart/ProductView currently send an empty-string currency field.

Sources/Helpers/Extension/ATTNSDK+Extension.swift
```diff
@@ -9,6 +9,91 @@ import Foundation
...
+            let orderTotal = products.reduce(0.0) {
+                $0 + (Double($1.price) ?? 0) * Double($1.quantity)
+            }
```
usharif-attentive: **🟠 Replace `Double(price)` math with `NSDecimalNumber`**

Two issues with the order-total math:

- `Double($1.price) ?? 0` silently zeroes out unparseable prices — producing a v2 Purchase event with a *wrong but plausible* total. At minimum, log a warning when parsing fails.
- `Double` for currency math introduces floating-point rounding (classic `0.1 + 0.2` territory).

`NSDecimalNumber` is already available on the source side (`item.price.price`). Sum decimals and format once at the end.

usharif-attentive (commented):
Ranked suggestions from the review — posted inline on the relevant lines.

Sources/API/ATTNEventTypes.swift
adelag-attentive: The type field on            
  ATTNMobileCustomEventMetadata is String? (optional), so existing call-sites
  that use sendCustomEvent(customProperties:) without a type will send null     
  which is fine for an additive optional field.

Tests/TestCases/ATTNSDKTests.swift
```diff
@@ -415,6 +415,138 @@ final class ATTNSDKTests: XCTestCase {
...
+        item1.quantity = 2
+        let item2 = ATTNItem(productId: "p2", productVariantId: "v2", price: ATTNPrice(price: NSDecimalNumber(string: "5.50"), currency: "USD"))
+        item2.quantity = 3
+        let order = ATTNOrder(orderId: "order-2")
+        let event = ATTNPurchaseEvent(items: [item1, item2], order: order)
+
+        sut.send(event: event)
+
+        XCTAssertTrue(apiSpy.sendNewEventWasCalled)
+        XCTAssertEqual(apiSpy.sendNewEventCallCount, 1)
```
usharif-attentive, line 459: **🟡 This test doesn't actually assert the total**

The name promises `computesCorrectTotal` and the inputs clearly expect `2 × 10.00 + 3 × 5.50 = 36.50`, but the body only asserts the API was called once — it never checks what `orderTotal` was sent.

This means the test would still pass if someone reverted the `NSDecimalNumber` fix back to `Double` math, defeating its purpose. Suggest capturing the metadata on `ATTNAPISpy` (similar to the existing `lastEventRequest`) and asserting the computed total:

```swift
// In ATTNAPISpy:
private(set) var lastPurchaseMetadata: ATTNPurchaseMetadata?
// ...in sendNewEvent, if metadata is ATTNPurchaseMetadata { lastPurchaseMetadata = $0 }

// In the test:
XCTAssertEqual(apiSpy.lastPurchaseMetadata?.orderTotal, "36.5")
```

(exact expected string depends on what `NSDecimalNumber.stringValue` produces — `"36.5"` or `"36.50"`.)

usharif-attentive (commented):
Follow-up on the purchase total test.

usharif-attentive (dismissed):

usharif-attentive (approved):

PR #257, opened 2026-05-13
MSDK-329: in-app SDK debug log overlay for Bonni

usharif-attentive:
## PR Type
- [x] Feature

## What's new?
Final piece of [MSDK-329](https://attentivemobile.atlassian.net/browse/MSDK-329) — an always-available, in-app SDK log viewer for the Bonni example app. Lets QA inspect SDK log output on TestFlight/release builds without attaching a debugger.

### SDK
- Wraps \`os.Logger\` with a new \`ATTNLogger\` struct that fans out to both the OS unified log AND an in-process ring buffer (\`ATTNLogBuffer\`). Existing 178 \`Loggers.creative.debug(...)\` / etc. call sites compile unchanged thanks to a custom \`ATTNLogMessage: ExpressibleByStringInterpolation\` that mirrors \`os.Logger\`'s \`, privacy:\` API.
- \`ATTNLogBuffer\` is a thread-safe ring buffer (1000 entries, amortized O(1) eviction) with a snapshot accessor and a multi-subscriber \`AsyncStream\`.
- New public API: \`ATTNSDK.recentLogs(since:)\` and \`ATTNSDK.logStream\`. Both work in any build configuration — including TestFlight — because the buffer captures \`.debug\` directly instead of relying on \`OSLogStore\`, which only persists \`.notice\` and above by default.
- New public type: \`ATTNLogEntry\` (\`@objc\`-bridged snapshot of one log line with a shared \`formatted(timestamp:)\` helper).

### Bonni
- New \`DebugLogOverlay\`: a separate \`UIWindow\` at \`windowLevel = .alert + 1\` installed once from \`SceneDelegate\` so it floats above every screen.
- A draggable \"LOGS\" toggle button shows/hides a compact, draggable floating panel (320×240) that streams entries live via \`ATTNSDK.logStream\` and renders them with timestamp, category, level, and message.
- The overlay window uses a \`hitTest\` override so non-chrome regions pass touches through to the underlying app — the rest of Bonni stays fully usable with the overlay installed.
- Panel work is skipped while hidden (entries queue to a small pending buffer and flush in one batch on toggle); visible appends use \`textStorage.append\` rather than reassigning \`textView.text\`, so per-entry cost stays O(1).

### Limitations / non-goals
- Privacy: every \`Loggers.*\` call is forwarded to \`os.Logger\` as a single \`.public\` value. Every existing call site already annotates \`.public\`, so this is a no-op today; documented inline so future contributors don't expect \`.private\` to redact in Console.app.
- Subscriber bookkeeping uses a UUID-keyed dict rather than reusing an extracted helper — there is no other \`AsyncStream\` subscriber-management code in the SDK to share with yet.

## Testing
- Builds the SDK framework and the Bonni app cleanly.
- Verified in the iOS Simulator: the floating LOGS button overlays the login screen with login UI fully tappable. Real SDK entries (\`network DEBUG: Success on attempt 1\`, \`event INFO: Obtained existing visitor id: ...\`, \`event DEBUG: Successfully sent push token\`) stream into the panel live.

## Screenshots
https://github.com/user-attachments/assets/bad66b06-cdbf-4191-8024-17c1cf6f9185

[MSDK-329]: https://attentivemobile.atlassian.net/browse/MSDK-329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Bonni/AttentiveExample/DebugLogOverlay.swift
+++ b/Bonni/AttentiveExample/DebugLogOverlay.swift
@@ -1,0 +1,268 @@
+//
+//  DebugLogOverlay.swift
+//  AttentiveExample
+//
+//  Created by Umair Sharif on 5/12/26.
+//
+
+import UIKit
+import ATTNSDKFramework
+
+/// A floating window that renders SDK log entries on top of any screen. Useful for
+/// inspecting logs in TestFlight/release builds without attaching a debugger.
+///
+/// - Tap the floating "LOGS" button to show/hide the panel.
+/// - Drag the LOGS button or the panel's title bar to reposition them.
+final class DebugLogOverlay {
+    static let shared = DebugLogOverlay()
+
+    private var window: OverlayWindow?
+    private var streamTask: Task<Void, Never>?
+
+    private init() {}
+
+    func install(on windowScene: UIWindowScene) {
+        guard window == nil else { return }
+        let window = OverlayWindow(windowScene: windowScene)
+        window.windowLevel = .alert + 1
+        window.isHidden = false
+        window.rootViewController = OverlayContainerViewController()
+        self.window = window
+
+        let containerVC = window.rootViewController as? OverlayContainerViewController
+        containerVC?.panel.replace(with: ATTNSDK.recentLogs())
+
+        let containerRef = containerVC
+        streamTask = Task {
+            for await entry in ATTNSDK.logStream {
+                await MainActor.run {
+                    containerRef?.panel.append(entry)
+                }
+            }
+        }
+    }
+}
+
+/// A window that ignores taps on its empty regions, so the underlying app keeps
+/// receiving touches everywhere except on the LOGS button and the floating panel.
+private final class OverlayWindow: UIWindow {
+    override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+        let view = super.hitTest(point, with: event)
+        if view === rootViewController?.view { return nil }
+        return view
+    }
+}
+
+private final class OverlayContainerViewController: UIViewController {
+    let panel = DebugLogPanelView(frame: CGRect(x: 16, y: 80, width: 320, height: 240))
+    private let toggleButton = UIButton(type: .system)
+    private var didPositionToggle = false
+
+    override func loadView() {
+        let root = UIView()
+        root.backgroundColor = .clear
+        view = root
+
+        toggleButton.frame = CGRect(x: 0, y: 0, width: 60, height: 44)
+        toggleButton.setTitle("LOGS", for: .normal)
+        toggleButton.titleLabel?.font = .systemFont(ofSize: 12, weight: .bold)
+        toggleButton.setTitleColor(.white, for: .normal)
+        toggleButton.backgroundColor = UIColor.black.withAlphaComponent(0.7)
+        toggleButton.layer.cornerRadius = 22
+        toggleButton.layer.borderColor = UIColor.white.withAlphaComponent(0.5).cgColor
+        toggleButton.layer.borderWidth = 1
+        toggleButton.addTarget(self, action: #selector(togglePanel), for: .touchUpInside)
+        toggleButton.addGestureRecognizer(UIPanGestureRecognizer(target: self, action: #selector(handleDrag(_:))))
+        root.addSubview(toggleButton)
+
+        panel.isHidden = true
+        root.addSubview(panel)
+    }
+
+    override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+        // Preserve user-dragged position across rotations.
+        guard !didPositionToggle else { return }
+        didPositionToggle = true
+        let bounds = view.bounds
+        let safeBottom = view.safeAreaInsets.bottom
+        toggleButton.frame.origin = CGPoint(x: bounds.width - 76, y: bounds.height - safeBottom - 144)
+    }
+
+    @objc private func togglePanel() {
+        let willShow = panel.isHidden
+        panel.isHidden = !willShow
+        if willShow { panel.flushPending() }
+    }
+
+    @objc private func handleDrag(_ gesture: UIPanGestureRecognizer) {
+        guard let target = gesture.view, let parent = target.superview else { return }
+        let translation = gesture.translation(in: parent)
+        target.center = CGPoint(x: target.center.x + translation.x, y: target.center.y + translation.y)
+        gesture.setTranslation(.zero, in: parent)
+    }
+}
+
+/// Compact, draggable floating log panel. The title bar is the drag handle.
+private final class DebugLogPanelView: UIView {
+    private let textView: UITextView = {
+        let tv = UITextView()
+        tv.font = .monospacedSystemFont(ofSize: 10, weight: .regular)
+        tv.isEditable = false
+        tv.backgroundColor = .clear
+        tv.textColor = .white
+        tv.alwaysBounceVertical = true
+        tv.translatesAutoresizingMaskIntoConstraints = false
+        return tv
+    }()
+
+    private let titleBar: UIView = {
+        let v = UIView()
+        v.backgroundColor = UIColor.white.withAlphaComponent(0.1)
+        v.translatesAutoresizingMaskIntoConstraints = false
+        return v
+    }()
+
+    private let titleLabel: UILabel = {
+        let l = UILabel()
+        l.text = "SDK Logs"
+        l.textColor = .white
+        l.font = .systemFont(ofSize: 11, weight: .semibold)
+        l.translatesAutoresizingMaskIntoConstraints = false
+        return l
+    }()
+
+    private let clearButton: UIButton = {
+        let b = UIButton(type: .system)
+        b.setTitle("Clear", for: .normal)
+        b.setTitleColor(.white, for: .normal)
+        b.titleLabel?.font = .systemFont(ofSize: 11, weight: .semibold)
+        b.translatesAutoresizingMaskIntoConstraints = false
+        return b
+    }()
+
+    private static let timestampFormatter: DateFormatter = {
+        let f = DateFormatter()
+        f.dateFormat = "HH:mm:ss.SSS"
+        return f
+    }()
+
+    /// Lines pending append when the panel is hidden. Flushed in one batch on show.
+    private var pendingLines: [String] = []
+    private var lineCount = 0
+    private static let maxLines = 1000
+
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        backgroundColor = UIColor.black.withAlphaComponent(0.85)
+        layer.cornerRadius = 8
+        layer.borderColor = UIColor.white.withAlphaComponent(0.4).cgColor
+        layer.borderWidth = 1
+        clipsToBounds = true
+
+        addSubview(titleBar)
+        titleBar.addSubview(titleLabel)
+        titleBar.addSubview(clearButton)
+        addSubview(textView)
+
+        clearButton.addTarget(self, action: #selector(clearTapped), for: .touchUpInside)
+        titleBar.addGestureRecognizer(UIPanGestureRecognizer(target: self, action: #selector(handleDrag(_:))))
+        titleBar.isUserInteractionEnabled = true
+
+        NSLayoutConstraint.activate([
+            titleBar.topAnchor.constraint(equalTo: topAnchor),
+            titleBar.leadingAnchor.constraint(equalTo: leadingAnchor),
+            titleBar.trailingAnchor.constraint(equalTo: trailingAnchor),
+            titleBar.heightAnchor.constraint(equalToConstant: 28),
+
+            titleLabel.leadingAnchor.constraint(equalTo: titleBar.leadingAnchor, constant: 10),
+            titleLabel.centerYAnchor.constraint(equalTo: titleBar.centerYAnchor),
+
+            clearButton.trailingAnchor.constraint(equalTo: titleBar.trailingAnchor, constant: -10),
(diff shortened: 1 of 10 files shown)


Sources/Helpers/ATTNLogBuffer.swift
```diff
@@ -0,0 +1,63 @@
...
+            guard let since else { return entries }
+            return entries.filter { $0.date >= since }
+        }
+    }
+
+    /// Async stream of new entries appended after subscription. The caller must
+    /// retain the resulting Task; cancelling it removes the subscription.
+    func stream() -> AsyncStream<ATTNLogEntry> {
+        AsyncStream { continuation in
+            let id = UUID()
```
adelag-attentive, line 68: `AsyncStream` without an explicit buffering policy defaults to unbounded buffering. If a subscriber consumes slower than logs are produced (UI hidden, Task suspended), yielded entries accumulate in memory without limit. Since `ATTNSDK.logStream` is public and multi-subscriber, one lagging consumer can grow unboundedly while others keep up.

Recommend specifying a bounded policy:
```swift
AsyncStream(bufferingPolicy: .bufferingNewest(100)) { continuation in ... }
```
This drops the oldest pending entries when a consumer falls behind — acceptable for a debug tool.

Sources/Public/ATTNLogEntry.swift
```diff
@@ -0,0 +1,48 @@
...
+@objc(ATTNLogEntry)
+public final class ATTNLogEntry: NSObject {
+    public enum Level: String, CaseIterable {
+        case debug
+        case info
+        case notice
+        case warning
+        case error
+        case fault
+    }
```
adelag-attentive, line 20: `Level` is defined as a typed enum, but the `level` property is stored as a raw `String`. Swift consumers lose compiler-checked exhaustiveness and must do manual string matching.

Presumably this is for `@objc` bridging. Consider adding a computed property for Swift callers:
```swift
public var logLevel: Level? { Level(rawValue: level) }
```
At minimum, document that `level` always contains one of the `Level.rawValue` strings.

Sources/Helpers/Logger.swift
```diff
@@ -8,10 +8,100 @@
...
+
+    func info(_ message: ATTNLogMessage) {
+        osLogger.info("\(message.rendered, privacy: .public)")
+        ATTNLogBuffer.shared.append(level: .info, category: category, message: message.rendered)
+    }
+
+    func notice(_ message: ATTNLogMessage) {
+        osLogger.notice("\(message.rendered, privacy: .public)")
+        ATTNLogBuffer.shared.append(level: .notice, category: category, message: message.rendered)
+    }
```
adelag-attentive: `os.Logger` defers string interpolation until the system decides to persist the message. On release builds, `.debug` entries are typically discarded before interpolation occurs — zero cost.

With `ATTNLogMessage`, interpolation is evaluated eagerly at every call site to populate `rendered`, regardless of whether `os.Logger` would have discarded it. On hot paths (e.g., debug logging on every network response), this overhead could be measurable.

Consider gating buffer appends behind a runtime flag (e.g., `ATTNLogBuffer.shared.isEnabled`) so builds that don't use the overlay pay nothing, or benchmark to confirm the cost is acceptable.

Bonni/AttentiveExample/DebugLogOverlay.swift
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    private func appendVisibleLine(_ line: String) {
+        let prefix = textView.text.isEmpty ? "" : "\n"
+        textView.textStorage.append(NSAttributedString(
+            string: prefix + line,
+            attributes: [
+                .font: textView.font ?? UIFont.monospacedSystemFont(ofSize: 10, weight: .regular),
+                .foregroundColor: textView.textColor ?? UIColor.white
+            ]
```
adelag-attentive: The panel uses `textStorage.append` for O(1) per-entry appends, but `trimIfNeeded()` walks the entire text from the start to locate newlines (O(n)), then reassigns `textView.text` — discarding the existing `textStorage` and rebuilding from scratch.

There is also a bookkeeping issue: `flushPending()` joins multiple entries with `"\n"` into a single string and calls `appendVisibleLine` once, incrementing `lineCount` by 1. The actual text contains multiple lines, so `lineCount` undercounts and trimming triggers later than expected.

Consider maintaining entries in a `[String]` array and rebuilding from the tail on trim, or using `textStorage.replaceCharacters(in:with:"")` to remove the leading range without full reassignment.

adelag-attentive (commented):
General comment: The ring buffer has non-trivial eviction logic (amortized `capacity + evictionSlack` boundary) and the subscriber lifecycle has edge cases (e.g., `onTermination` firing concurrently with appends). Core SDK additions like this should have unit test coverage — particularly the eviction boundary and subscriber cleanup path.

Sources/Helpers/Logger.swift
adelag-attentive, line 72: **Eager interpolation cost on release builds**

With `os.Logger`, `.debug` messages on release builds are essentially free — the system discards them *before* evaluating the string interpolation closure. The interpolation is never executed, so there's zero CPU or allocation cost at the call site.

`ATTNLogMessage` changes this contract. Because it conforms to `ExpressibleByStringInterpolation`, the compiler materializes the full `ATTNLogMessage` value (including evaluating every interpolation segment into `rendered`) *at the call site*, before `debug(_:)` is even entered. So by the time your function body runs and `ATTNLogBuffer.shared.append()` checks `isCapturing`, the string has already been built.

Concretely, a call like:
```swift
Loggers.creative.debug("Resizing creative to \(newArea.width)x\(newArea.height)")
```
…previously cost **nothing** on a release build (os.Logger discards `.debug` without interpolating). Now it unconditionally allocates a String and runs `String.append` for each segment — even when `isCapturing = false` and the message is thrown away.

For infrequent log sites this doesn't matter, but the SDK has debug logging on network callbacks, creative resize events, and navigation decisions — paths that can fire many times per second.

**Options:**
1. Accept the cost and document it (if benchmarking shows it's negligible for our call sites).
2. Make `debug(_:)` accept `@autoclosure () -> ATTNLogMessage` so interpolation is deferred until you confirm the buffer wants it. You'd still forward to `os.Logger` with its native interpolation separately (preserving its lazy semantics), and only evaluate the autoclosure when `isCapturing` is true.
3. Use a compile-time flag to strip buffer appends entirely from production builds that don't ship the overlay.

I'd lean toward at least benchmarking before merging to confirm the overhead is acceptable for the hot paths.

Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -68,6 +68,27 @@ public final class ATTNSDK: NSObject {
...
+        get { ATTNLogBuffer.shared.isCapturing }
+        set { ATTNLogBuffer.shared.isCapturing = newValue }
+    }
+
+    /// Recent SDK log entries from the in-process ring buffer. Returns an empty
+    /// array when log capture is disabled (the default).
+    /// - Parameter since: Optional cutoff; entries older than this are filtered out.
+    public static func recentLogs(since: Date? = nil) -> [ATTNLogEntry] {
+        ATTNLogBuffer.shared.snapshot(since: since)
+    }
```
adelag-attentive: **Missing `@objc` on new public APIs**

Per the SDK's coding standards, all public API surface should have `@objc` annotations for Objective-C consumers. `isLogCaptureEnabled` (Bool) and `recentLogs(since:)` (returns `[ATTNLogEntry]` where `ATTNLogEntry` is an NSObject subclass) are both ObjC-bridgeable and should get `@objc`.

`logStream` can't be bridged (AsyncStream isn't representable in ObjC), so that one is fine without it.

adelag-attentive (commented):
Two review comments — one on interpolation cost, one on @objc bridging.

adelag-attentive (commented):
## Review Comments

### 1. `@autoclosure` + `@escaping` in `ATTNLogger` — still evaluates on the `else` branch (Medium)

```swift
func debug(_ message: @autoclosure @escaping () -> ATTNLogMessage) {
    if ATTNLogBuffer.shared.isCapturing {
        let rendered = message().rendered
        osLogger.debug("\(rendered, privacy: .public)")
        ...
    } else {
        osLogger.debug("\(message().rendered, privacy: .public)")  // ← evaluates here too
    }
}
```

The `else` branch still eagerly evaluates `message()` to produce the string for `os.Logger`. The comment says "os.Logger keeps its native gating" but that's not quite right — `os.Logger.debug` would normally not evaluate its `OSLogMessage` argument on release builds (it's compiler-optimized away). Here, you're pre-rendering into a `String` and passing it as `"\(rendered, privacy: .public)"`, which forces evaluation regardless of whether unified logging would have dropped it.

**The net effect:** when `isCapturing = false` and on a release build, every `Loggers.*.debug(...)` call still allocates a `String` via `ATTNLogMessage` interpolation. This is a regression from the original `os.Logger` direct usage where `.debug` was truly zero-cost on release.

**Suggestion:** Consider a fast-path: if `!isCapturing` AND the level is `.debug`, skip entirely (don't even call `osLogger.debug`). Or, only forward to `osLogger` using its native string interpolation (not pre-rendered strings) when capture is off — though this is hard to do generically.

This is the one thing I'd want addressed before merge, since the SDK has ~178 `.debug` call sites that now all allocate on every invocation in production.

### 2. Overlay window may steal key window status (Medium — potential UI issue)

```swift
window.windowLevel = .alert + 1
window.isHidden = false
```

The overlay window is made visible immediately during `scene(_:willConnectTo:)` — the same method where the app's main window calls `makeKeyAndVisible()`. Since the overlay window is created *after* `makeKeyAndVisible()`, UIKit may promote it to the key window. If it does:
- Text fields in the app beneath might not receive keyboard focus properly
- `UIApplication.shared.keyWindow` (deprecated but still used by some code/third-party libraries) could return the overlay window
- Any SDK code or third-party library that uses `keyWindow` to present alerts, action sheets, or web views could attach them to the overlay window instead of the app window

**Fix:** Override `canBecomeKey` on `OverlayWindow` to return `false`:

```swift
override var canBecomeKey: Bool { false }
```

This ensures the overlay never steals key window status regardless of window level or creation order.

### 3. No bounds checking on drag — button/panel can be dragged off-screen with no recovery (Medium — UX issue)

Both the LOGS button and the panel use an unclamped drag handler:

```swift
@objc private func handleDrag(_ gesture: UIPanGestureRecognizer) {
    guard let target = gesture.view, let parent = target.superview else { return }
    let translation = gesture.translation(in: parent)
    target.center = CGPoint(x: target.center.x + translation.x, y: target.center.y + translation.y)
    gesture.setTranslation(.zero, in: parent)
}
```

If the user drags the LOGS button fully off-screen, there's no way to get it back — no reset mechanism exists. On device rotation, the button stays at its dragged position (`didPositionToggle = true` after first layout), so if it was off-screen in portrait it stays off-screen in landscape.

**Suggestion:** Clamp the center so at least 20px of the view remains visible within the parent's bounds. Or add a double-tap-to-reset gesture on the button.

adelag-attentive:
shipping this to unblock inbox e2e testing -
can you add a toggle button (if it's not already there just in case) and then go ahead and ship that PR?

adelag-attentive (dismissed):
per slack thread https://attentivemobile.slack.com/archives/C09PRH450MC/p1784663381587279?thread_ts=1784662494.358519&cid=C09PRH450MC

usharif-attentive:
Rebased onto `develop` and retargeted the PR base from `main` to `develop` (the two branches had diverged; the diff is now just this feature's commits). Status of review feedback:

**Already addressed in earlier follow-up commits:**
- **Eager interpolation on release builds** — `defer log interpolation when buffer capture is off`: each level method takes the message as an `@autoclosure` and, when the buffer isn't capturing, forwards it into `os.Logger`'s own lazy interpolation, so disabled `.debug` calls cost nothing. Capture is also now opt-in via `ATTNSDK.isLogCaptureEnabled` (off by default; Bonni enables it at launch).
- **Unbounded `AsyncStream` buffering** — `stream()` now uses `.bufferingNewest(capacity)`, so a lagging subscriber drops oldest pending entries instead of growing without bound.
- **`ATTNLogEntry.level` stringly-typed** — added `logLevel: Level?` typed accessor and documented the invariant on `level`.
- **Missing `@objc`** — `isLogCaptureEnabled` and `recentLogs(since:)` are `@objc`; `logStream` can't bridge.
- **Panel trim/lineCount bookkeeping** — panel now keeps a `lines: [String]` source of truth with `renderedCount`/`pendingSplice`, splices leading lines in place instead of rebuilding `textView.text`, and no longer undercounts multi-line flushes.
- **Ring buffer test coverage** — `ATTNLogBufferTests` covers the eviction boundary and subscriber add/terminate lifecycle.

**New in this push:**
- **Key window stealing** — `OverlayWindow` overrides `canBecomeKey` to return `false`.
- **Drag off-screen with no recovery** — drags on the LOGS button and panel are clamped so at least 24pt stays on-screen, and positions re-clamp on rotation.
- Moved the debug-log API into an `ATTNSDK` extension — the rebase pushed the class body past SwiftLint's 700-line `type_body_length` error threshold.

Verified: full test suite passes on iPhone 17 sim (228 passed, 0 failed) and Bonni builds cleanly.

usharif-attentive:
Added a **"Show SDK Log Overlay"** switch to Bonni's Settings screen (`4372beb`) — the overlay chrome (LOGS button included) can now be hidden entirely. The preference persists in `UserDefaults`, applies live, and defaults to **visible** so QA/TestFlight behavior is unchanged. Log capture keeps running while hidden, so re-enabling the overlay still shows full history. Verified both states in the simulator.

PR #258, opened 2026-05-19
MSDK-312: protect mutable shared state with NSLock

usharif-attentive:
## Summary

[MSDK-312](https://attentivemobile.atlassian.net/browse/MSDK-312) called out unprotected mutable state across `ATTNSDK`, `ATTNAPI`, `ATTNUserIdentity`, and `ATTNEventTracker` that can race under real-world conditions (app backgrounding, push arrivals, SDK init). This PR adds the `NSLock`-based synchronization the ticket recommended.

- `ATTNUserIdentity` — `identifiers` and `visitorId` are now backed by lock-guarded storage; `mergeIdentifiers` is atomic, fixing the read-modify-write TOCTOU
- `ATTNEventTracker` — lock around `_sharedInstance` setup/get/destroy
- `ATTNAPI` — lock around `lastPushTokenSendTime` debounce check-and-set
- `ATTNSDK` — lock guarding `pendingURL` and `lastRegularOpenTime` (`pendingMarketingRequests` was already serialized via `marketingQueue`, left unchanged)
- `cachedGeoAdjustedDomain` from the ticket no longer applies — removed in MSDK-264

Second commit cleans up the lock sites with an `NSLock.withLock` helper (RAII unlock on every exit path, including the early-return debounce paths) and consolidates the duplicated `DispatchGroup` test boilerplate behind an `XCTestCase.runConcurrently(iterations:queueLabels:)` helper.

## Test plan
- [x] `bundle exec fastlane ios unit_test` passes locally
- [x] New concurrency tests exercise the protected state under concurrent readers/writers (`ATTNUserIdentityTests.testMergeIdentifiers_concurrentMerges_doesNotCrashAndPreservesAllKeys`, `testMergeAndRead_concurrentReadersAndWriters_doesNotCrash`, `testClearUser_concurrentClearAndMerge_doesNotCrash`, `ATTNEventTrackerTests.testSharedInstance_concurrentSetupAndAccess_doesNotCrash`)
- [ ] CI green

[MSDK-312]: https://attentivemobile.atlassian.net/browse/MSDK-312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/API/ATTNAPI.swift
+++ b/Sources/API/ATTNAPI.swift
@@ -18,7 +18,7 @@ final class ATTNAPI: ATTNAPIProtocol {
 
     private let retryClient: ATTNRetryingNetworkClient
     private let stateLock = NSLock()
-    private var _lastPushTokenSendTime: Date?
+    private var lastPushTokenSendTime: Date?
 
     // MARK: ATTNAPIProtocol Properties
     var domain: String
@@ -62,10 +62,10 @@ final class ATTNAPI: ATTNAPIProtocol {
         // debounce to remove duplicate events; only allow events tracking at most once every 2 seconds
         let shouldSkip: Bool = stateLock.withLock {
             let now = Date()
-            if let last = _lastPushTokenSendTime, now.timeIntervalSince(last) < 2 {
+            if let last = lastPushTokenSendTime, now.timeIntervalSince(last) < 2 {
                 return true
             }
-            _lastPushTokenSendTime = now
+            lastPushTokenSendTime = now
             return false
         }
         if shouldSkip {
--- a/Sources/API/ATTNAPI.swift
+++ b/Sources/API/ATTNAPI.swift
@@ -61,14 +61,17 @@ final class ATTNAPI: ATTNAPIProtocol {
                                          callback: ATTNAPICallback?) {
         // debounce to remove duplicate events; only allow events tracking at most once every 2 seconds
         let now = Date()
-        stateLock.lock()
-        if let last = _lastPushTokenSendTime, now.timeIntervalSince(last) < 2 {
-            stateLock.unlock()
+        let shouldSkip: Bool = stateLock.withLock {
+            if let last = _lastPushTokenSendTime, now.timeIntervalSince(last) < 2 {
+                return true
+            }
+            _lastPushTokenSendTime = now
+            return false
+        }
+        if shouldSkip {
             Loggers.event.debug("Skipping duplicate sendPushToken due to debounce.")
             return
         }
-        _lastPushTokenSendTime = now
-        stateLock.unlock()
 
         Loggers.network.debug("Sending push token - Visitor ID: \(userIdentity.visitorId, privacy: .public), Push Token: \(pushToken, privacy: .public), Auth Status: \(authorizationStatus.rawValue, privacy: .public)")
 
--- a/Sources/API/ATTNAPI.swift
+++ b/Sources/API/ATTNAPI.swift
@@ -17,7 +17,8 @@ final class ATTNAPI: ATTNAPIProtocol {
     private(set) var urlSession: URLSession
 
     private let retryClient: ATTNRetryingNetworkClient
-    private var lastPushTokenSendTime: Date?
+    private let stateLock = NSLock()
+    private var _lastPushTokenSendTime: Date?
 
     // MARK: ATTNAPIProtocol Properties
     var domain: String
@@ -60,11 +61,14 @@ final class ATTNAPI: ATTNAPIProtocol {
                                          callback: ATTNAPICallback?) {
         // debounce to remove duplicate events; only allow events tracking at most once every 2 seconds
         let now = Date()
-        if let last = lastPushTokenSendTime, now.timeIntervalSince(last) < 2 {
+        stateLock.lock()
+        if let last = _lastPushTokenSendTime, now.timeIntervalSince(last) < 2 {
+            stateLock.unlock()
             Loggers.event.debug("Skipping duplicate sendPushToken due to debounce.")
             return
         }
-        lastPushTokenSendTime = now
+        _lastPushTokenSendTime = now
+        stateLock.unlock()
 
         Loggers.network.debug("Sending push token - Visitor ID: \(userIdentity.visitorId, privacy: .public), Push Token: \(pushToken, privacy: .public), Auth Status: \(authorizationStatus.rawValue, privacy: .public)")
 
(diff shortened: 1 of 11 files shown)


adelag-attentive (commented):
## Review Comments

### 1. `ATTNUserIdentity.identifiers` — verify callers handle snapshot semantics

The `identifiers` getter now returns a value-type copy under the lock. This changes observable behavior: callers that previously held a reference to the stored property saw live mutations; now they get a frozen snapshot. Since `identifiers` is the most critical piece of data in the SDK — it drives identity resolution, event attribution, and opt-in/out payloads — this warrants an audit of call sites to confirm none rely on the old live-reference behavior. If any caller reads `identifiers` and later expects it to reflect a concurrent `mergeIdentifiers` call without re-reading, that's now silently broken. Worth grepping for uses and verifying each one is a read-and-discard pattern.

### 2. `ATTNEventTracker.setup` — log after lock release has subtle ordering

```swift
sharedInstanceLock.withLock {
    _sharedInstance = ATTNEventTracker(sdk: sdk)
}
Loggers.event.debug("ATTNEventTracker was initialized with SDK")
```

If two threads race `setup`, Thread A sets the instance, Thread B overwrites it, then Thread A logs "was initialized" — implying A's instance is live when B's already replaced it. Moving the log inside the lock (or logging the instance pointer) would make debug output more accurate.

### 3. `Date()` captured before lock in debounce (ATTNAPI + ATTNSDK)

```swift
let now = Date()
let shouldSkip: Bool = stateLock.withLock {
    if let last = _lastPushTokenSendTime, now.timeIntervalSince(last) < 2 { ... }
```

Under contention, a thread could block on the lock while `now` ages. Two threads capturing nearly-identical timestamps before either acquires the lock could both pass the debounce. For a 2-second window this is negligible, but if the window ever tightens, `Date()` should move inside the lock.

### 4. Concurrency tests — consider asserting invariants, not just crash-freedom

`testMergeIdentifiers_concurrentMerges_doesNotCrashAndPreservesAllKeys` asserts `XCTAssertNotNil` but doesn't verify no keys were dropped. Since each iteration merges a unique key (`key0`..`key199`), you could assert the final `customIdentifiers` dictionary has exactly 200 entries — that would prove atomicity, not just crash-freedom. Similarly, there's no test for concurrent `consumeDeepLink` + deep-link-setting, which would prove only one caller gets the URL from the read-and-clear.

usharif-attentive:
Addressed all feedback in 3e7ec12:

**Codex P1 / Adela #1 — `identifiers` setter removed.** Restored the public setter; it validates and replaces the dictionary atomically under the lock, preserving the original behavior (validation skipped for empty dicts, willSet semantics). Obj-C `setIdentifiers:` is back.

**Adela #2 — log outside lock.** Moved `Loggers.event.debug("ATTNEventTracker was initialized with SDK")` inside the `sharedInstanceLock.withLock` block so racing setups can't print misleading ordering.

**Adela #3 — `Date()` outside lock.** Moved `let now = Date()` inside the locked region in both `ATTNAPI.sendPushToken` and `ATTNSDK.handleRegularOpen`. No longer possible for two threads to both observe a stale `now` and both pass the debounce.

**Adela #4 — weak test assertions.**
- Renamed `testMergeIdentifiers_concurrentMerges_doesNotCrashAndPreservesAllKeys` → `testMergeIdentifiers_concurrentMerges_preservesAllKeys`. Each iteration now writes a *unique* top-level key (`dynamic_key_0`..`dynamic_key_199`) and the test asserts `identifiers.count == 200` — proves atomicity, not just crash-freedom. The original test merged everything under the same `customIdentifiers` key so 200 entries was unreachable.
- Added `testConsumeDeepLink_concurrentConsumeAndSet_atMostOneConsumerWinsPerSet` in `ATTNSDKTests` to exercise the `consumeDeepLink` read-and-clear under concurrent setters. Exposed `normalizeAndBroadcast` as `internal` so the test can drive it.

Branch was also rebased onto current main. All unit tests pass locally.

adelag-attentive:
looks good after the codex comment is addressed!

usharif-attentive:
> looks good after the codex comment is addressed!

@adelag-attentive the codex comment was addressed in this commit: https://github.com/attentive-mobile/attentive-ios-sdk/commit/3e7ec1288d79511080617cb1a590ea03c48400e1

adelag-attentive (dismissed):

adelag-attentive (dismissed):

adelag-attentive (commented):
## `clearUser()` concurrent calls — verify UserDefaults consistency
(very last comment, everything else looks good!)
Two threads calling `clearUser()` simultaneously will each call `createNewVisitorId()` (which persists to UserDefaults), but only the last thread to acquire the lock actually sets `_visitorId`. If `getVisitorId()` on next app launch reads from UserDefaults rather than the in-memory value, you could get a stale/mismatched visitor ID.

In practice this is unlikely (nobody calls `clearUser` from two threads simultaneously), but visitorID is the most important piece of identifier in the SDK — it drives identity resolution, event attribution, and opt-in/out payloads — so we need to be safe rather than sorry. Worth confirming the read path on next launch can't diverge from the in-memory value.

usharif-attentive:
> `clearUser()` concurrent calls — verify UserDefaults consistency

Good catch — addressed in 5fd0103. You're right that hoisting `createNewVisitorId()` out of the lock decoupled the UserDefaults write order from the in-memory swap order:

```
Thread A: createNewVisitorId() → saves "AAA" to UserDefaults
Thread B: createNewVisitorId() → saves "BBB" to UserDefaults
Thread A: lock { _visitorId = "AAA" }
Thread B: lock { _visitorId = "BBB" }
```

With the writes possibly hitting disk in either order, the next app launch (which reads from UserDefaults via `getVisitorId()`) could see `"AAA"` while the last in-memory value was `"BBB"`. Reverted to doing generate+save+swap all under the lock so the disk write order matches the in-memory order. The earlier clear-concurrency test timeout is already addressed by capping that test at 20 iterations.

Sources/API/ATTNAPI.swift
```diff
@@ -17,7 +17,8 @@ final class ATTNAPI: ATTNAPIProtocol {
     private(set) var urlSession: URLSession
 
     private let retryClient: ATTNRetryingNetworkClient
-    private var lastPushTokenSendTime: Date?
+    private let stateLock = NSLock()
+    private var _lastPushTokenSendTime: Date?
```
adelag-attentive: we're almost there - very last issue, i'm not sure why this pr renamed a bunch of variables by adding an underscore. i know it was an objective c thing, but do we actually need it here? if possible, should we add some guidelines to the style guide for claude?

adelag-attentive (dismissed):

adelag-attentive (approved):

PR #266, opened 2026-06-05
MSDK-377: wire real endpoint for inbox unread count

adelag-attentive:
## Summary
- Adds async `fetchInboxUnreadCount` on `ATTNAPI` that calls `POST /inbox/messages/unread/count` per the Inbox RFC, with `visitor_id` / `push_token` / `email` / `phone` in the payload.
- `InboxManager.unreadCount` is now **server-authoritative** — replaces the prior locally-derived value (computed from `messagesByID.values.filter { !$0.isRead }`). Local `markRead`/`markUnread` no longer mutate it; future mark-read API responses will (separate ticket).
- Exposes new public `ATTNSDK.refreshInboxUnreadCount()` so host apps can refresh on app launch / push open per RFC guidance.
- `InboxManager.init` now fans out the mock-message load and the unread-count fetch concurrently via `async let`.

## Jira
https://attentivemobile.atlassian.net/browse/MSDK-377

## Out of scope (future tickets)
- The `/inbox/messages` list fetch — still uses the existing mock generator.
- `mark-read`, `mark-unread`, `delete`, `track-click` endpoints.
- Centralizing identity-forwarding for additional inbox endpoints (will revisit when the 2nd inbox endpoint lands).

## Test plan
- [x] New `ATTNInboxAPITests` (6 tests) — success, body shape, omitted-empty fields, server error, malformed JSON, network error
- [x] `ATTNAPITests` and `ATTNSDKTests` still pass
- [ ] Manual: hit the real endpoint from Bonni once the backend is deployed; verify `await sdk.unreadCount` returns the server value after `refreshInboxUnreadCount()`

## Notes for reviewers
- Base branch is **develop** (not main) — Inbox is an experimental feature behind that branch.
- Heads up: `develop` is currently behind `main` by ~30 commits (MSDK-371/370/312/354/315/344, etc.) — separate ticket to sync.

### DIFF
--- a/Sources/Inbox/InboxManager.swift
+++ b/Sources/Inbox/InboxManager.swift
@@ -13,9 +13,17 @@ public enum InboxState: Sendable {
     case error(Error)
 }
 
+/// Identifiers needed to call the inbox endpoints.
+struct InboxIdentitySnapshot: Sendable {
+    let visitorId: String
+    let pushToken: String
+    let email: String?
+    let phone: String?
+}
+
 /// Closure that supplies the identifiers needed to call the inbox endpoints.
 /// Resolved at call-time so the manager always uses the latest values from `ATTNSDK`.
-typealias InboxIdentityProvider = @Sendable () -> (userIdentity: ATTNUserIdentity, pushToken: String, email: String?, phone: String?)
+typealias InboxIdentityProvider = @Sendable () -> InboxIdentitySnapshot
 
 actor InboxManager {
     private var messagesByID: [Message.ID: Message] = [:]
@@ -25,11 +33,16 @@ actor InboxManager {
 
     /// Server-authoritative unread count. Updated only by `refreshUnreadCount()` and (future)
     /// mark-read/mark-unread/delete API responses — local `markRead`/`markUnread` mutations
-    /// do not change this value.
+    /// do not change this value. Defaults to 0 until the first successful refresh, so callers
+    /// cannot distinguish "no unread" from "not yet fetched".
     private(set) var unreadCount: Int = 0
 
-    private let api: ATTNAPIProtocol?
-    private let identityProvider: InboxIdentityProvider?
+    private let api: ATTNAPIProtocol
+    private let identityProvider: InboxIdentityProvider
+
+    /// Monotonic counter used to discard responses from refresh calls that have been
+    /// superseded by a newer in-flight call.
+    private var refreshGeneration: UInt = 0
 
     var allMessages: [Message] {
         if let cached = cachedSortedMessages {
@@ -60,13 +73,13 @@ actor InboxManager {
         }
     }
 
-    init(api: ATTNAPIProtocol? = nil, identityProvider: InboxIdentityProvider? = nil) {
+    init(api: ATTNAPIProtocol, identityProvider: @escaping InboxIdentityProvider) {
         self.api = api
         self.identityProvider = identityProvider
+        // `currentState` is already `.loading`; kick off the mock-message load and the
+        // unread-count fetch concurrently. Both run on this actor so they're serialized
+        // at write boundaries — the parallelism is for the awaits, not the state mutations.
         Task {
-            await send(.loading)
-            // Run mock-message load and unread-count fetch concurrently — they touch
-            // independent state and the network call should not wait on the local mock.
             async let messages: Void = updateMessages(Self.getMockInbox().messages)
             async let count: Void = refreshUnreadCount()
             _ = await (messages, count)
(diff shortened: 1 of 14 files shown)


Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -51,7 +51,20 @@ public final class ATTNSDK: NSObject {
...
+                return InboxIdentitySnapshot(visitorId: "", pushToken: "", email: nil, phone: nil)
+            }
+            let identifiers = self.userIdentity.identifiers
+            return InboxIdentitySnapshot(
+                visitorId: self.userIdentity.visitorId,
+                pushToken: self.currentPushToken,
+                email: identifiers[ATTNIdentifierType.email] as? String,
+                phone: identifiers[ATTNIdentifierType.phone] as? String
+            )
+        })
```
usharif-attentive: **Concurrency: data race on `userIdentity.identifiers`**

This closure is declared `@Sendable`, but it captures `self` (`ATTNSDK`) and reads `self.userIdentity.identifiers` / `self.userIdentity.visitorId` from the `InboxManager` actor's executor. Meanwhile, `identify(_:)` → `userIdentity.mergeIdentifiers(...)` and `clearUserIdentifiers()` mutate that same `[String: Any]` from the caller's thread (typically main).

Concurrent read+write on a Swift `Dictionary` is a data race — possible torn reads, lost writes, or a trap. TSan would flag this. The window is narrow today (only fires if `identify`/`clearUser` overlaps an in-flight refresh) but it's a real bug worth a follow-up before this graduates from `develop` → `main`. Snapshot identifiers under a serial queue, or have the writer publish an immutable snapshot the provider reads atomically.

Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -259,13 +272,24 @@ public final class ATTNSDK: NSObject {
...
         get async {
             await inboxManager.unreadCount
         }
     }
 
+    /// Refreshes the unread inbox message count from the server.
+    /// Per RFC, host apps should call this when navigating to the app's main page (e.g. on app launch)
+    /// and after opening a push notification.
+    public func refreshInboxUnreadCount() async {
+        await inboxManager.refreshUnreadCount()
```
usharif-attentive: **Stale unread count after identity switch**

This new public surface returns a cached count that survives identity changes. `clearUser()` and `updateUser(...)` both call `clearUserIdentifiers()` (which generates a new visitor ID) but don't touch `InboxManager.unreadCount`. After they return, `await sdk.unreadCount` keeps reporting the previous user's count until the host app explicitly calls `refreshInboxUnreadCount()`.

Small privacy/UX leak — a logged-out account's unread badge stays visible to the next user on the device. Probably fine to defer (the PR already scopes mark-read API wiring out), but worth a ticket: on identity change, reset the cached count to 0 and/or kick off an immediate refresh.

Sources/Inbox/InboxManager.swift
```diff
@@ -77,8 +134,19 @@ actor InboxManager {
         send(.loaded(allMessages))
     }
 
-    func refresh() async {
-        await updateMessages(Self.getMockInbox().messages)
+    func refresh() {
```
usharif-attentive: **Blocker:** `refresh()` no longer refreshes the unread count.

Before this PR, `unreadCount` was locally derived from `messagesByID`, so reloading messages implicitly refreshed it. Now that the count is server-authoritative, `refresh()` only reloads mock messages — pull-to-refresh in `InboxView` (via `InboxViewModel.refresh()`) will not re-fetch `/inbox/messages/unread/count`.

A user who reads a message on another device won't see the badge update via pull-to-refresh — only on app launch or push open. Given the whole point of this PR is server-authoritative counts, that's a functional gap.

Suggested fix: call `refreshUnreadCount()` from `refresh()` (concurrently with the message reload once the real list endpoint lands), or explicitly document why pull-to-refresh is intentionally excluded.

Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -216,6 +223,9 @@ public final class ATTNSDK: NSObject {
     private func clearUserIdentifiers() {
         let oldVisitorId = userIdentity.visitorId
         userIdentity.clearUser()
+        publishIdentitySnapshot()
+        // Drop the previous user's cached unread count so the next user doesn't see their badge.
+        Task { [inboxManager] in await inboxManager.resetUnreadCount() }
```
usharif-attentive: **Blocker:** this forces lazy-init of `inboxManager` on every `clearUser()`, even for host apps that never use inbox.

Resolving `self.inboxManager` here triggers the `lazy var`'s initializer, whose `Task` runs `updateMessages(getMockInbox())` **and** `refreshUnreadCount()`. Net effect: a host app that has never touched any inbox API will now issue a network call to `/inbox/messages/unread/count` at logout time — a behavior change for all existing consumers on `develop`.

Two options:
1. Store `inboxManager` as an `Optional` (or wrap in a small "materialized" flag) and only call `resetUnreadCount()` if it's already been created.
2. Drop the init-time `refreshUnreadCount()` fetch and require an explicit `refreshInboxUnreadCount()` from the host — the public API already exposes it per the RFC.

Option 2 also fixes the subtler issue that the init-time fetch races the very first `identify()` call (it can fire against a freshly-generated anonymous visitor ID before the host has identified the user).

usharif-attentive (commented):
Two blocking items — details inline. Rest of the earlier review is polish / follow-up ticket material.

Sources/Inbox/InboxManager.swift
```diff
@@ -13,13 +13,63 @@ public enum InboxState: Sendable {
...
+    /// Monotonic counter used to discard responses from refresh calls that have been
+    /// superseded by a newer in-flight call or an identity reset.
+    private var refreshGeneration: UInt = 0
+
     var allMessages: [Message] {
+        get async {
+            await awaitInitialRefresh()
+            return sortedMessagesSnapshot()
+        }
+    }
```
usharif-attentive, line 55: Question on intent: `init` now only fires `performUnreadCountFetch` — it no longer loads messages. `messagesByID` is populated exclusively by `updateMessages`, which is only reached via `refresh()` (called from `InboxView.task`/`.refreshable`).

So for a host that reads `await sdk.allMessages` (or subscribes to `inboxStateStream`) **without presenting `inboxView()`**:
- `allMessages` returns `[]`
- `currentState` stays `.loading` indefinitely (no `send(.loaded)` is reachable without a message load), so the stream emits `.loading` and never advances.

The old `init()` eagerly called `updateMessages(getMockInbox().messages)`, so both surfaces used to resolve. Is deferring message loading to view presentation intended (list fetch is out of scope / mock data / experimental)? If so, worth updating the `allMessages` doc comment, which currently says first access "triggers a background unread-count fetch" but doesn't mention messages stay empty. If passive consumption is a real use case, this is a regression — `allMessages` should trigger a load and `.loading` needs a resolution path.

adelag-attentive: good call, it'll be easier once the fetch all messages PR lands

usharif-attentive (approved):

PR #267, opened 2026-06-15
MSDK-390: Add AGENTS.md for AI agent SDK integration

usharif-attentive:
## Ticket
[MSDK-390](https://attentivemobile.atlassian.net/browse/MSDK-390) — iOS counterpart to [MSDK-349](https://attentivemobile.atlassian.net/browse/MSDK-349) (Android, [PR #219](https://github.com/attentive-mobile/attentive-android-sdk/pull/219))

## Summary
- Adds `AGENTS.md` at the repo root as a tightly-scoped integration guide for AI coding agents (Claude Code, Cursor, Copilot, Codex, etc.)
- Functions as an alternative to reading the full `README.md` — focused on agent execution, not human reading
- Scope is intentionally limited to the minimum viable integration: dependency wiring (SPM, CocoaPods, or manual XCFramework), `ATTNSDK.initialize` + `ATTNEventTracker.setup` in the host `AppDelegate`, and an interactive push-setup flow gated on the user opting into push
- Explicitly forbids the agent from wiring `identify`/`clearUser`/`updateUser`, recording events, rendering creatives, or wiring marketing subscriptions, so client agents don't over-implement on the first pass
- Adds a "Setup with an AI coding agent" section to `README.md` that points to `AGENTS.md` with a copy-pastable trigger prompt

## Typical client usage

A client developer integrating Attentive into their iOS app points their AI agent at this repo (or copies `AGENTS.md` into their own repo) and gives a one-line prompt:

> "Set up the Attentive iOS SDK in this app."

The agent then:

1. **Inspects the client codebase** — detects dependency manager (Podfile vs SPM artifacts), Swift vs Objective-C, app lifecycle style (UIKit `AppDelegate`, UIKit + `SceneDelegate`, or SwiftUI `App`), deployment target, and any existing `UNUserNotificationCenterDelegate`.
2. **Confirms the dependency manager** with the user before writing — the detected default is pre-selected, but the user can override.
3. **Asks for the Attentive domain** (and defaults `mode` to `.debug` for first-time integration, with a note to switch to `.production` for release).
4. **Adds the dependency** at the latest version (looked up from GitHub releases at integration time, not hardcoded).
5. **Wires init in `AppDelegate.application(_:didFinishLaunchingWithOptions:)`** — either editing the existing class or, for pure-SwiftUI apps, asking before adding a `UIApplicationDelegateAdaptor`. Matches the host's language.
6. **Asks whether push is wanted.** If no, sets `pushEnabled: false` and stops. If yes, walks an interactive sub-flow: APNs entitlement detection (instructs the user to add the capability rather than editing entitlements directly), `UNUserNotificationCenterDelegate` hooks (appends to existing handlers rather than replacing them), optional Notification Service Extension for image attachments, and deep-link routing via `.ATTNSDKDeepLinkReceived` or `consumeDeepLink()`.
7. **Stops.** Builds to verify, then tells the user to replace the domain placeholder if needed and points them to `README.md` for identify, events, creatives, marketing subscriptions, etc.

The "Things NOT to do" section keeps the agent from drifting into deleting existing push handlers, calling `UIApplication.shared.registerForRemoteNotifications()` (the SDK already does that), bumping the deployment target, or introducing DI frameworks.

## Test plan
- [ ] Verify `AGENTS.md` renders correctly on GitHub
- [ ] Dry-run the guide against a sample iOS app (UIKit `AppDelegate` + CocoaPods) and confirm the agent produces a working integration
- [ ] Dry-run against a SwiftUI `App` with no `AppDelegate` and confirm the agent stops and asks before adding one
- [ ] Dry-run with SPM and confirm dependency confirmation flow works
- [ ] Confirm the agent does not touch identify/events/creatives/marketing-subscriptions when given the base prompt

[MSDK-390]: https://attentivemobile.atlassian.net/browse/MSDK-390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MSDK-349]: https://attentivemobile.atlassian.net/browse/MSDK-349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -35,6 +35,22 @@ Do not invent a domain. Always initialize the SDK in `.debug` mode for first-tim
 
 ---
 
+## Step 0 — Confirm the host builds clean
+
+Before touching anything, confirm the host produces a clean baseline build. This prevents you from mis-attributing pre-existing host issues to the SDK integration later.
+
+1. Look for a project-specific bootstrap script or setup command in the repo's `README`, `Makefile`, `Justfile`, `bin/`, or `scripts/` directory (e.g. `bin/setup`, `make bootstrap`, `bundle install && pod install`). If one exists, run it.
+2. Build the host with its standard command — e.g. `xcodebuild -workspace <…>.xcworkspace -scheme <…> -destination 'generic/platform=iOS Simulator' build`, or whatever the project documents.
+3. If the baseline build fails, **stop** and tell the user:
+
+   > "I tried to build the host before integrating Attentive and the build failed with `<error>`. This looks unrelated to the SDK — can you fix the host first (or confirm it's expected and you want me to proceed anyway)? I don't want to attribute pre-existing failures to the integration."
+
+   Do not attempt to debug pre-existing host issues as part of this integration.
+
+Only proceed to Step 1 once you have a clean baseline build (or the user has explicitly told you to continue anyway).
+
+---
+
 ## Step 1 — Inspect the client codebase
 
 Before editing anything, determine:
@@ -80,19 +96,31 @@ If the lookup fails (no network, rate-limited, etc.), tell the user and ask them
 
 Use the resolved version (e.g. `2.0.15`) in place of `<VERSION>` below.
 
+> **Note on version drift:** the version Xcode's resolver actually pins can differ from what you looked up (a stale package cache or an existing `Package.resolved` will hold an older version, even with a `from:` constraint). After installing — Step 2c — you must read back the *resolved* version and report **that** to the user, not the latest-tag lookup result. Step 2c spells out where to read it.
+
 ### 2b. Add the dependency
 
 #### Swift Package Manager (recommended)
 
-If the host uses SPM, add the package via Xcode's UI **or** edit the project's `Package.swift` / Xcode-managed package list. The remote URL is:
+For most iOS app hosts, SPM is managed through Xcode's GUI — package references live inside `project.pbxproj`, not in an editable app-level `Package.swift` — so the package add is a **user-driven step you cannot perform yourself**. Treat this as the primary path; the `Package.swift` variant below is only for the smaller case where the host *is* an SPM package.
 
-```
-https://github.com/attentive-mobile/attentive-ios-sdk
-```
+The package coordinates:
 
-Pin to `<VERSION>` (exact) or `from: "<VERSION>"` (semver-compatible — recommended). The product name to link against the app target is **`ATTNSDKFramework`**.
+- **URL:** `https://github.com/attentive-mobile/attentive-ios-sdk`
+- **Version rule:** `Up to Next Major Version` from `<VERSION>` (semver-compatible — recommended), or `Exact` `<VERSION>` if the host prefers strict pinning.
+- **Product to link:** `ATTNSDKFramework`
 
-If the host has a `Package.swift` (i.e. the app itself is an SPM package), add to its dependencies:
+##### Primary path — Xcode GUI handoff
+
+Tell the user:
+
+> "Open the project in Xcode → File → Add Package Dependencies… → paste `https://github.com/attentive-mobile/attentive-ios-sdk` → choose `Up to Next Major Version` from `<VERSION>` → add `ATTNSDKFramework` to your app target. Let me know once it resolves and I'll wire up the init."
+
+Wait for the user to confirm the package resolved before continuing. You cannot add an SPM package by editing files for an Xcode-managed app target.
+
+##### Variant — host *is* an SPM package
+
+Only use this path if you found a top-level `Package.swift` in the repo (i.e. the host is itself a Swift package, not an Xcode app project). Add to its dependencies:
 
 ```swift
 .package(url: "https://github.com/attentive-mobile/attentive-ios-sdk", from: "<VERSION>")
@@ -103,10 +131,6 @@ If the host has a `Package.swift` (i.e. the app itself is an SPM package), add t
 ```swift
 .product(name: "ATTNSDKFramework", package: "attentive-ios-sdk")
 ```
-
-If you cannot edit `Package.swift` (e.g. SPM is managed via Xcode's GUI), tell the user:
-
-> "Open the project in Xcode → File → Add Package Dependencies… → paste `https://github.com/attentive-mobile/attentive-ios-sdk` → choose `Up to Next Major Version` from `<VERSION>` → add `ATTNSDKFramework` to your app target. Let me know once it resolves and I'll wire up the init."
 
 #### CocoaPods
 
@@ -132,7 +156,7 @@ After install, the user must open the `.xcworkspace` (not the `.xcodeproj`).
 
 Only fall back to manual XCFramework integration if the user explicitly requests it or both SPM and CocoaPods are unavailable. The README has the full instructions; warn the user that incorrect embedding can cause App Store submission failures.
 
-### 2c. Verify the dependency resolves
+### 2c. Verify the dependency resolves, and read back the resolved version
 
 Before writing any code that imports `ATTNSDKFramework`, make sure the project builds:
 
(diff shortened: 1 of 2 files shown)


adelag-attentive:
excited for this! I'm going to pull an open source project and try it on later today/tmrw morning

AGENTS.md
```diff
@@ -0,0 +1,521 @@
...
+
+…and to the relevant target's `dependencies`:
+
+```swift
+.product(name: "ATTNSDKFramework", package: "attentive-ios-sdk")
+```
+
+If you cannot edit `Package.swift` (e.g. SPM is managed via Xcode's GUI), tell the user:
+
+> "Open the project in Xcode → File → Add Package Dependencies… → paste `https://github.com/attentive-mobile/attentive-ios-sdk` → choose `Up to Next Major Version` from `<VERSION>` → add `ATTNSDKFramework` to your app target. Let me know once it resolves and I'll wire up the init."
```
adelag-attentive: **Quick fix #1 — reframe Xcode-GUI handoff as the primary path, not a fallback.**

The section currently reads "If you cannot edit `Package.swift` (e.g. SPM is managed via Xcode's GUI), tell the user…" — i.e. GUI handoff is treated as exception-handling. In practice, on the dry-run I just did against a large host, this was the *only* path: the host's app-target SPM was Xcode-GUI-managed (Package references live in `project.pbxproj`, no editable app-level `Package.swift`). Suspect this is the common case for most hosts an agent would be pointed at, not the exception.

Suggest swapping section order so the GUI handoff is presented first and the `Package.swift` edit is the variant for hosts where the app *is* an SPM package. The agent should *expect* to halt and instruct the user, not treat it as a fallback.

_(PR author can verify whether this matches their experience across the apps they've tested — my sample size is one host, so this assumption may not hold across CocoaPods-heavy hosts or SDK-only consumers.)_

AGENTS.md
```diff
@@ -0,0 +1,521 @@
...
+   > "Do you know your Attentive domain? It's the short identifier for your account (e.g. `myshop`)."
+
+   - If the user says **yes**, immediately follow up with: "What is it?" Wait for their answer and use that exact string in the config. Do not proceed until they've given you the domain.
+   - If the user says **no** (or doesn't know), insert `"YOUR_ATTENTIVE_DOMAIN"` as a placeholder and tell them to replace it before shipping.
+
+Do not invent a domain. Always initialize the SDK in `.debug` mode for first-time integration; tell the user to switch to `.production` for release builds.
+
+---
+
+## Step 1 — Inspect the client codebase
```
adelag-attentive, line 54: **Quick fix #2 — add a Step 0 pre-flight: "verify the host builds clean *before* you start."**

On my dry-run the host was in a half-set-up state: a project-specific bootstrap script hadn't been run (a generated build dependency was missing) and there were two pre-existing Swift compile errors unrelated to Attentive that only surfaced under a newer Xcode. The build failed *during* Step 4 — and an agent following the guide can't easily distinguish "my edits broke it" from "the host was already broken."

Suggested addition before Step 1:

> **Step 0 — Confirm the host builds clean.** Before touching anything, run the host's documented build/bootstrap command (`xcodebuild build`, `bundle exec fastlane …`, project README's setup steps, etc.) and confirm it produces a clean baseline build. If it doesn't, stop and ask the user to fix or skip the broken host — do not attempt to debug pre-existing host issues as part of this integration.

This prevents the agent from incorrectly attributing host build failures to the SDK integration, and gives the user one clean sentence to react to.

AGENTS.md
```diff
@@ -0,0 +1,521 @@
...
+
+## Step 4 — Verify the base build
+
+1. Build the app (⌘B, or `xcodebuild -workspace <…>.xcworkspace -scheme <…> -destination 'generic/platform=iOS Simulator' build`).
+2. If the build fails, fix it before moving on. Common issues:
+   - **`No such module 'ATTNSDKFramework'`** — SPM/CocoaPods didn't resolve. Re-run `pod install` or have Xcode resolve packages.
+   - **CocoaPods + Xcode opens the wrong file** — make sure you're working in `<…>.xcworkspace`, not `<…>.xcodeproj`.
+3. Once the build succeeds, proceed to Step 5.
+
+Do not run the app on a simulator or device unless asked.
```
adelag-attentive: **Quick fix #3 — add an explicit runtime verification at the end of Step 4.**

"Build succeeded" doesn't mean the SDK initialized at runtime. The init code's failure branch is `print("Attentive SDK failed to initialize: \(error)")` — silent unless someone reads the console. So an agent that stops at "build green" hands off without ever knowing whether the SDK actually came up.

On my dry-run, the only signal that the integration actually worked was tailing the simulator log:

```
Initializing ATTNSDK v2.0.14, Mode: debug, Domain: <...>, PushEnabled: true
ATTNSDK initialization successful - Visitor ID: <...>
Successfully sent event of type 'i'
```

That last line — the network round-trip to `events.attentivemobile.com` succeeding — is the only confirmation of "working" vs. "compiles."

Suggested addition as Step 4.5:

> **Step 4.5 — Confirm the SDK initializes at runtime.** After a successful build, launch the app on a simulator and tail the system log:
>
> ```bash
> xcrun simctl spawn booted log stream --level debug --predicate 'eventMessage CONTAINS "ATTNSDK"'
> ```
>
> Within a few seconds of launch you should see `Initializing ATTNSDK v…` followed by `ATTNSDK initialization successful`. If you only see `Attentive SDK failed to initialize:`, surface that error to the user before continuing.

This is also the cheapest test the agent can run on its own without the user.

AGENTS.md
```diff
@@ -0,0 +1,521 @@
...
+        _ application: UIApplication,
+        didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil
+    ) -> Bool {
+        ATTNSDK.initialize(domain: "YOUR_ATTENTIVE_DOMAIN", mode: .debug) { result in
+            switch result {
+            case .success(let sdk):
+                self.attentiveSdk = sdk
+                ATTNEventTracker.setup(with: sdk)
+            case .failure(let error):
+                print("Attentive SDK failed to initialize: \(error)")
```
adelag-attentive, line 219: **Quick fix #4 — replace `print(...)` with a TODO pointing at the host's logger.**

Hard-coded `print("Attentive SDK failed to initialize: \(error)")` is fine for sample code, but most hosts of any size have a structured logger (`os.Logger`, `OSLog`, `swift-log`, custom wrappers) and dropping a `print` call into `AppDelegate.didFinishLaunchingWithOptions` is a small regression in those hosts.

Suggested change to the snippet (and the equivalent at lines 220 and 336):

```swift
case .failure(let error):
    // TODO(attentive): route to the host's logger if one exists.
    print("Attentive SDK failed to initialize: \(error)")
```

Cheap, costs nothing if ignored, gives the agent (or user) a clear breadcrumb to swap in `logger.log(...)` / `os_log(...)` / whatever the host uses.

AGENTS.md
```diff
@@ -0,0 +1,521 @@
...
+   - **UIKit + SceneDelegate**: `AppDelegate` plus a `SceneDelegate.swift`. SDK init still goes in `AppDelegate.application(_:didFinishLaunchingWithOptions:)` — push and lifecycle hooks for Attentive are still on the `AppDelegate`, not the scene delegate.
+   - **SwiftUI `App`** (`@main struct MyApp: App`): the project may not have an `AppDelegate`. In that case, you'll need to add one via `@UIApplicationDelegateAdaptor` — see Step 3.
+5. **Deployment target**: Note the value from the app target's `IPHONEOS_DEPLOYMENT_TARGET`. The SDK requires **iOS 14.0+**. The `Inbox` module additionally requires iOS 15.0+ (you are not wiring Inbox in this pass, so 14.0 is sufficient). If the host's deployment target is below 14.0, **stop and ask the user** before raising it — bumping the deployment target is their decision.
+6. **Existing `UNUserNotificationCenterDelegate`**: search for `UNUserNotificationCenterDelegate` in the project. If one exists already, you'll add Attentive's hooks to its callbacks rather than installing a new delegate.
+
+---
+
+## Step 2 — Add the dependency
+
+### 2a. Look up the latest SDK version
```
adelag-attentive, line 83: **Quick fix #5 — note the version-resolution gotcha and tell the agent to verify what was *actually* pinned.**

The guide says "Pin to `<VERSION>` (exact) or `from: "<VERSION>"` (semver-compatible — recommended)." On my dry-run the lookup at integration time returned `2.0.15` and that's what I told Xcode to use, but Xcode's resolver actually pulled `2.0.14` at runtime (visible in the init log). Likely a cache/index state thing, but the gap means the version the agent reports to the user ("installed 2.0.15") and the version that ships are not the same.

Two options:

1. After resolution, have the agent read back the resolved version (from `Package.resolved` or `pod install` output) and report *that* number to the user, not the latest-tag lookup result.
2. For first-time integrations, recommend `exact:` over `from:` so the resolved version matches the latest-tag lookup unambiguously.

Either would make the version the user thinks they have match the version that's actually pinned.

adelag-attentive (commented):
Dry-ran the no-push base integration end-to-end against a real client iOS app (large UIKit + SPM-via-Xcode-GUI host on iOS 26.1). The integration *worked* — app launched, `ATTNSDK.initialize` succeeded, first event delivered to `events.attentivemobile.com`. Verdict: **B+**. The Swift the guide tells the agent to write is correct, the inspection rubric in Step 1 is the strongest part, and the `var attentiveSdk: ATTNSDK?` + closure-init shape drops cleanly into a 400-line existing AppDelegate without fighting the host's lifecycle. Six suggestions below to tighten the parts that go fuzzy at the boundary between agent autonomy and Xcode/user actions. None are blockers — the guide as written produces a working integration.

AGENTS.md
adelag-attentive: **Quick fix #6 — "Do not run the app on a simulator or device unless asked" conflicts with #3.**

The guide says this twice (here and at line 497). If you adopt the runtime-verification idea from #3, this needs a carve-out, e.g.:

> Do not run the app on a simulator or device for demo or interactive purposes — but you may briefly launch it from the simulator log to confirm SDK initialization, then terminate it.

Mostly a wording fix, but worth doing in the same pass so the agent doesn't read the stricter rule and skip the verification step.

adelag-attentive (approved):

PR #271, opened 2026-07-08
MSDK-312: stub UserDefaults in concurrent clearUser test

usharif-attentive:
## Summary
- `testClearUser_concurrentClearAndMerge_doesNotCrash` was timing out on CircleCI because `clearUser()` synchronously writes a new visitor id through `UserDefaults.standard` while holding the identity lock; the previous mitigation (`c7942f7`) — bump timeout to 20s at 20 iterations — was still tripping on CI.
- Split `ATTNUserIdentity.init(identifiers:)` into a public `@objc` convenience init plus an internal designated init that accepts an `ATTNVisitorService`, so tests can inject alternate storage. Public API is unchanged (same `initWithIdentifiers:` selector, still `NSObject`, still `@objc`).
- Added `ATTNPersistentStorageMock` (in-memory, lock-protected) and rewired the failing test to inject it via `ATTNVisitorService(persistentStorage:)`. The test now asserts the lock, not disk I/O — runs 200 iterations at the default 5s timeout instead of 20 at 20s.

## Public API impact
None. `ATTNUserIdentity.init(identifiers:)` remains the only `@objc` init exposed to Swift/ObjC consumers. The new designated init is internal (no access modifier → module-private) and non-`@objc`, so it isn't visible outside the framework.

## Verification
- iPhone 16 simulator (iOS 26.1) — `xcodebuild test -only-testing:"attentive-ios-sdk Tests/ATTNUserIdentityTests"` → all 14 pass; the previously failing test now completes in **0.017s** (was timing out at 20s on CI).
- Same simulator — `ATTNVisitorServiceTests` and `ATTNUserIdentityExtensionTests` still pass (touched shared init/protocol surface).

## Host-app rollback
Revert this commit. No release-branch or migration coupling; only test/plumbing changes and a non-public `ATTNUserIdentity` init overload.

### DIFF
--- a/Sources/Public/ATTNUserIdentity.swift
+++ b/Sources/Public/ATTNUserIdentity.swift
@@ -34,10 +34,14 @@ public final class ATTNUserIdentity: NSObject {
     }
 
     @objc(initWithIdentifiers:)
-    public init(identifiers: [String: Any]) {
-        self.visitorService = .init()
+    public convenience init(identifiers: [String: Any]) {
+        self.init(identifiers: identifiers, visitorService: .init())
+    }
+
+    init(identifiers: [String: Any], visitorService: ATTNVisitorService) {
+        self.visitorService = visitorService
         self._identifiers = identifiers
-        self._visitorId = self.visitorService.getVisitorId()
+        self._visitorId = visitorService.getVisitorId()
         super.init()
     }
 
--- a/Tests/Doubles/Mocks/ATTNPersistentStorageMock.swift
+++ b/Tests/Doubles/Mocks/ATTNPersistentStorageMock.swift
@@ -1,0 +1,24 @@
+//
+//  ATTNPersistentStorageMock.swift
+//  attentive-ios-sdk Tests
+//
+
+import Foundation
+@testable import ATTNSDKFramework
+
+final class ATTNPersistentStorageMock: ATTNPersistentStorageProtocol {
+    private let lock = NSLock()
+    private var storage: [String: AnyObject] = [:]
+
+    func save(_ value: AnyObject, forKey key: String) {
+        lock.withLock { storage[key] = value }
+    }
+
+    func readString(forKey key: String) -> String? {
+        lock.withLock { storage[key] as? String }
+    }
+
+    func delete(forKey key: String) {
+        lock.withLock { _ = storage.removeValue(forKey: key) }
+    }
+}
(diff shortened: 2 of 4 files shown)


adelag-attentive:
### 1. `attentive-ios-sdk.xcodeproj/project.pbxproj` — non-standard Xcode object IDs (25 chars, not 24)

The four new lines for `ATTNPersistentStorageMock.swift` use these IDs:

- `A9D1C316A0001000000000001` (PBXBuildFile)
- `A9D1C316A0001000000000002` (PBXFileReference)

Both are **25 hex characters**. Every other object ID in this pbxproj is exactly **24** — e.g. `FB2983E42C0F73990039759C`, `FB080C7F2C1CC9E500834BAA`. That's Xcode's 96-bit UUID format (96 bits ÷ 4 bits/hex = 24 chars).

#### How this likely happened

This is the fingerprint of a pbxproj entry that was **hand-written / tool-written** rather than generated by Xcode itself. Xcode assigns IDs by allocating a 96-bit value (usually random-with-a-timestamp-prefix — that's why real IDs cluster by creation time: `FB2983…` files were added in one session, `FB080C…` in another). When a human or a script adds a file entry directly by editing the pbxproj text, they have to invent an unused ID, and it's very easy to miscount the hex digits — especially when constructing something like `<prefix><counter>` and padding by hand. Here the ID looks like `A9D1C316A0` + `001000000000001` — a 10-char prefix + 15-char counter = 25, off by one.

The dead giveaways:
1. **Length is wrong by exactly one character** — not a typo of a hex digit, an off-by-one in padding.
2. **The two new IDs differ only in the last character** (`…001` vs `…002`) — Xcode-generated IDs for the same file (build-file + file-ref) never share prefixes; they're allocated independently from the same random pool.
3. **The prefix `A9D1C316A0` doesn't match any of the file's existing prefix clusters** (`FB…`, `F9…`) — so it wasn't cloned from a neighbor either.

#### Why it matters

Xcode is fairly forgiving on read — it'll open the project fine — but the first time someone edits project settings in the IDE, Xcode rewrites the pbxproj and may:
- normalize/renumber these IDs, producing a large unrelated diff on an unrelated PR, or
- (rarer) fail to round-trip cleanly if the ID collides with something in its allocator's space.

More practically: the malformed IDs make the file no longer round-trip-stable, which fights git history hygiene going forward.

#### Fix

Regenerate the entry by adding `ATTNPersistentStorageMock.swift` through Xcode itself:

1. Remove the four hand-written lines from `project.pbxproj`.
2. In Xcode, right-click the `Tests/Doubles/Mocks` group → **Add Files to \"attentive-ios-sdk\"…** → select the file, add to the `attentive-ios-sdk Tests` target.
3. Commit the resulting pbxproj diff — Xcode will assign two proper 24-char IDs.

Small effort, avoids a future \"why did Xcode rewrite half our project file\" surprise on someone else's PR.

usharif-attentive:
@adelag-attentive addressed your comment

adelag-attentive (approved):

PR #273, opened 2026-07-09
MSDK-410: wire real endpoint for mark message read

usharif-attentive:
## Summary

Wires `markRead` to the real `PATCH /inbox/messages/read` endpoint, replacing the previous local-only mutation. Part of the inbox real-endpoint rollout ([MSDK-410](https://attentivemobile.atlassian.net/browse/MSDK-410)).

## Key Changes

- `ATTNAPI.markMessagesRead(pushToken:visitorId:messageIds:)` issues `PATCH /inbox/messages/read` with the identifier-based body `{ visitor_id, push_token, message_ids }` and decodes `UpdateReadStatusResponse` (`{ messages: [{ message_id, is_read }], unread_count }`).
- `InboxManager.markRead` is now `async`: it optimistically flips `isRead` locally for immediate UI feedback, then reconciles per-message read status **and** the server-authoritative `unread_count` from the response. On failure it reverts the local flip so the UI reflects true state.
- Introduces the shared `UpdateReadStatusResponse` model (used by mark-unread in the stacked follow-up).
- The mark-read/unread response `unread_count` is treated as authoritative — no follow-up `refreshInboxUnreadCount` call, which would be a redundant round trip and could race a read replica into overwriting the fresh value with a stale one.

## Public API Impact

`ATTNSDK.markRead(for:)` was already `public` and `async`; its signature is unchanged and all call sites already `await`. No source/binary break for host apps. Behavior changes from local-only to a network call.

## Verification

- All 211 unit tests pass via `xcodebuild test` on the iOS 26.5 simulator (iPhone 17 Pro). Adds 6 API-level tests (`ATTNInboxMarkReadAPITests`: URL, PATCH method, body shape, empty push token, server-error, decode-error, network-error) and 4 `InboxManagerTests` cases (success reconciles `unread_count`, failure reverts the flip, already-read still calls the API, unknown id no-ops).
- `bundle exec fastlane ios lint` passes (0 serious violations).
- Exercised end-to-end in Bonni: swiping mark-read fires `PATCH /inbox/messages/read`, returns 200, and the SDK correctly decodes/stores/emits. **Note:** the prod backend currently routes to `MockInboxMessageClient`, which hardcodes `unread_count = 5` in every response, so the badge count does not move yet. The per-message flip is echoed correctly by the server. Badge movement will work once the real `InboxMessageClient` is swapped in server-side (raised with the backend team).

## Risk

Medium. Contained to the inbox mark-read path. The optimistic-flip + revert keeps the UI truthful even when the network fails. No change to identity, events, or push handling. Recent backend flakiness on the inbox endpoints (intermittent 401s in prod) means the call may fail in the field — but a failed call reverts cleanly and logs via `Loggers.network`.

## Observability

`Loggers.network` OSLog entries (`PATCH /inbox/messages/read`, status code, `Failed to mark inbox message read`) on-device; `mobile-api` service logs (`Received mark messages read request`) server-side. No dedicated dashboard for this endpoint yet.

## Rollback

No feature flag or config kill-switch — mitigation is a follow-up SDK release plus host-app re-adoption. The client-side revert-on-failure limits blast radius: a failing endpoint leaves the UI showing true read state rather than a wrong optimistic one.

[MSDK-410]: https://attentivemobile.atlassian.net/browse/MSDK-410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/API/ATTNAPI.swift
+++ b/Sources/API/ATTNAPI.swift
@@ -490,6 +490,58 @@ final class ATTNAPI: ATTNAPIProtocol {
         }
     }
 
+    /// Marks the supplied messages as read on the server.
+    ///
+    /// PATCH /inbox/messages/read — identifier-based (visitor_id + push_token), body carries
+    /// the list of message ids. The response echoes the per-message read status and the
+    /// updated unread count so the caller can reconcile local state without a follow-up fetch.
+    func markMessagesRead(
+        pushToken: String,
+        visitorId: String,
+        messageIds: [String]
+    ) async throws -> UpdateReadStatusResponse {
+        Loggers.network.debug("Marking inbox messages read - Visitor ID: \(visitorId, privacy: .public), Push Token: \(pushToken, privacy: .public), Count: \(messageIds.count, privacy: .public)")
+
+        var payload: [String: Any] = [
+            "visitor_id": visitorId,
+            "message_ids": messageIds
+        ]
+        if !pushToken.isEmpty { payload["push_token"] = pushToken }
+
+        guard let url = URL(string: "https://mobile.attentivemobile.com/inbox/messages/read") else {
+            Loggers.network.error("Invalid inbox mark-read URL")
+            throw ATTNError.badURL
+        }
+
+        var request = URLRequest(url: url)
+        request.httpMethod = "PATCH"
+        request.timeoutInterval = 15
+        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+        request.setValue("1", forHTTPHeaderField: "x-datadog-sampling-priority")
+        request.httpBody = try JSONSerialization.data(withJSONObject: payload, options: [])
+
+        Loggers.network.debug("PATCH /inbox/messages/read")
+
+        let (data, response) = try await dataTask(with: request)
+
+        guard let http = response as? HTTPURLResponse else {
+            Loggers.network.error("Inbox mark-read returned a non-HTTP response")
+            throw ATTNError.inboxResponseDecodeFailed
+        }
+        Loggers.network.debug("Inbox mark-read status code: \(http.statusCode, privacy: .public)")
+        guard (200..<300).contains(http.statusCode) else {
+            Loggers.network.error("Inbox mark-read API returned status \(http.statusCode, privacy: .public)")
+            throw ATTNError.inboxRequestFailed(statusCode: http.statusCode)
+        }
+
+        do {
+            return try JSONDecoder().decode(UpdateReadStatusResponse.self, from: data)
+        } catch {
+            Loggers.network.error("Failed to decode inbox mark-read response: \(error.localizedDescription, privacy: .public)")
+            throw ATTNError.inboxResponseDecodeFailed
+        }
+    }
+
     private static let inboxJSONDecoder: JSONDecoder = {
         let decoder = JSONDecoder()
         // Accept ISO-8601 timestamps with or without fractional seconds. Foundation's built-in
--- a/Sources/API/ATTNAPI.swift
+++ b/Sources/API/ATTNAPI.swift
@@ -502,11 +502,11 @@ final class ATTNAPI: ATTNAPIProtocol {
     ) async throws -> UpdateReadStatusResponse {
         Loggers.network.debug("Marking inbox messages read - Visitor ID: \(visitorId, privacy: .public), Push Token: \(pushToken, privacy: .public), Count: \(messageIds.count, privacy: .public)")
 
-        var payload: [String: Any] = [
-            "visitor_id": visitorId,
-            "message_ids": messageIds
-        ]
-        if !pushToken.isEmpty { payload["push_token"] = pushToken }
+        let body = UpdateReadStatusRequest(
+            visitorId: visitorId,
+            pushToken: Self.inboxPushToken(pushToken),
+            messageIds: messageIds
+        )
 
         guard let url = URL(string: "https://mobile.attentivemobile.com/inbox/messages/read") else {
             Loggers.network.error("Invalid inbox mark-read URL")
@@ -518,7 +518,7 @@ final class ATTNAPI: ATTNAPIProtocol {
         request.timeoutInterval = 15
         request.setValue("application/json", forHTTPHeaderField: "Content-Type")
         request.setValue("1", forHTTPHeaderField: "x-datadog-sampling-priority")
-        request.httpBody = try JSONSerialization.data(withJSONObject: payload, options: [])
+        request.httpBody = try JSONEncoder().encode(body)
 
         Loggers.network.debug("PATCH /inbox/messages/read")
 
@@ -540,6 +540,13 @@ final class ATTNAPI: ATTNAPIProtocol {
             Loggers.network.error("Failed to decode inbox mark-read response: \(error.localizedDescription, privacy: .public)")
             throw ATTNError.inboxResponseDecodeFailed
         }
+    }
+
+    /// The inbox backend expects push tokens namespaced by transport (e.g. `apns:<token>`).
+    /// Returns `nil` for an empty token so `push_token` is omitted from the request body
+    /// rather than sent as an empty (and unusable) identifier.
+    private static func inboxPushToken(_ pushToken: String) -> String? {
+        pushToken.isEmpty ? nil : "apns:\(pushToken)"
     }
 
     private static let inboxJSONDecoder: JSONDecoder = {
--- a/Sources/Inbox/InboxManager.swift
+++ b/Sources/Inbox/InboxManager.swift
@@ -52,9 +52,10 @@ actor InboxManager {
     /// load returning inside the refresh window resets that flag but must not release this one.
     private var isRefreshingFirstPage: Bool = false
 
-    /// Server-authoritative unread count. Only `refreshUnreadCount()` writes it; local
-    /// `markRead` / `markUnread` do not. Reads should go through the `unreadCount` accessor,
-    /// which awaits the init-time refresh so the first read never returns a stale 0.
+    /// Server-authoritative unread count. Written by `refreshUnreadCount()` and by the mark-read
+    /// API response (see `applyReadStatusResponse`); local `markUnread` does not touch it. Reads
+    /// should go through the `unreadCount` accessor, which awaits the init-time refresh so the
+    /// first read never returns a stale 0.
     private var storedUnreadCount: Int = 0
 
     private let api: ATTNAPIProtocol
@@ -332,10 +333,35 @@ actor InboxManager {
         }
     }
 
-    func markRead(_ messageID: Message.ID) {
-        guard messagesByID[messageID]?.isRead == false else { return }
-        messagesByID[messageID]?.isRead = true
-        send(.loaded(orderedMessagesSnapshot()))
+    /// Marks a message read. Optimistically flips the local state, then reconciles from the
+    /// server response. On failure the local flip is reverted so the UI reflects true state.
+    func markRead(_ messageID: Message.ID) async {
+        guard let previousIsRead = messagesByID[messageID]?.isRead else { return }
+        if !previousIsRead {
+            messagesByID[messageID]?.isRead = true
+            send(.loaded(allMessages))
+        }
+
+        let identity = identityProvider()
+        guard !identity.visitorId.isEmpty else {
+            Loggers.network.debug("Skipping inbox mark-read: empty visitor ID")
+            return
+        }
+
+        do {
+            let response = try await api.markMessagesRead(
+                pushToken: identity.pushToken,
+                visitorId: identity.visitorId,
+                messageIds: [messageID]
+            )
+            applyReadStatusResponse(response)
+        } catch {
+            Loggers.network.error("Failed to mark inbox message read: \(error.localizedDescription, privacy: .public)")
+            if messagesByID[messageID] != nil, !previousIsRead {
+                messagesByID[messageID]?.isRead = previousIsRead
+                send(.loaded(allMessages))
+            }
+        }
     }
 
     func markUnread(_ messageID: Message.ID) {
@@ -417,5 +443,15 @@ extension InboxManager {
             messagesByID[message.id] = message
             messageOrder.append(message.id)
         }
+    }
+
+    /// Reconcile per-message read status and the authoritative unread count from a
+    /// mark-read/mark-unread response.
+    private func applyReadStatusResponse(_ response: UpdateReadStatusResponse) {
+        for status in response.messages {
+            messagesByID[status.messageId]?.isRead = status.isRead
+        }
+        unreadCount = response.unreadCount
+        send(.loaded(allMessages))
     }
 }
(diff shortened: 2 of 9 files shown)


usharif-attentive:
@codex review

usharif-attentive:
@attentive-claude review

adelag-attentive:
### 2. `Sources/API/ATTNAPI.swift:511` — Manual JSON encoding via `JSONSerialization` when a `Codable` model would be cleaner

The request body is currently built as `[String: Any]` with a conditional key:

```swift
var payload: [String: Any] = [
    "visitor_id": visitorId,
    "message_ids": messageIds
]
if !pushToken.isEmpty { payload["push_token"] = pushToken }
// ...
request.httpBody = try JSONSerialization.data(withJSONObject: payload, options: [])
```

The rest of this codebase is `Codable` on the response side (`UpdateReadStatusResponse`, `InboxResponse`, etc.). Mixing `JSONSerialization` on the way out and `JSONDecoder` on the way back is a small but real consistency snag — and the `[String: Any]` payload gives up compile-time checking on the field names / types (e.g. `"visitor_id"` misspellings, `messageIds` accidentally becoming `[Int]`, etc.).

Consider a tiny request struct:

```swift
private struct MarkMessagesReadRequest: Encodable {
    let visitorId: String
    let pushToken: String?
    let messageIds: [String]

    enum CodingKeys: String, CodingKey {
        case visitorId = "visitor_id"
        case pushToken = "push_token"
        case messageIds = "message_ids"
    }
}

let body = MarkMessagesReadRequest(
    visitorId: visitorId,
    pushToken: pushToken.isEmpty ? nil : pushToken,
    messageIds: messageIds
)
request.httpBody = try JSONEncoder().encode(body)
```

`encodeIfPresent` on the optional `pushToken` gives you the "omit when empty" behavior for free, and the wire contract now lives next to the code that produces it.

Non-blocking — the current form works and the test coverage pins the wire shape. Worth folding in here (base of the stack) so the mark-unread follow-up in #274 inherits the pattern rather than duplicating the `JSONSerialization` shape.

Sources/API/ATTNAPI.swift
```diff
@@ -504,6 +504,58 @@ final class ATTNAPI: ATTNAPIProtocol {
...
+        visitorId: String,
+        messageIds: [String]
+    ) async throws -> UpdateReadStatusResponse {
+        Loggers.network.debug("Marking inbox messages read - Visitor ID: \(visitorId, privacy: .public), Push Token: \(pushToken, privacy: .public), Count: \(messageIds.count, privacy: .public)")
+
+        var payload: [String: Any] = [
+            "visitor_id": visitorId,
+            "message_ids": messageIds
+        ]
+        if !pushToken.isEmpty { payload["push_token"] = pushToken }
```
adelag-attentive: From the latest back end team updates: please prepend "apns:" to all push tokens for all inbox related requests

Sources/Inbox/InboxManager.swift
```diff
@@ -211,9 +210,35 @@ actor InboxManager {
...
+    /// Marks a message read. Optimistically flips the local state, then reconciles from the
+    /// server response. On failure the local flip is reverted so the UI reflects true state.
+    func markRead(_ messageID: Message.ID) async {
+        guard let previousIsRead = messagesByID[messageID]?.isRead else { return }
+        if !previousIsRead {
+            messagesByID[messageID]?.isRead = true
+            send(.loaded(allMessages))
+        }
+
+        let identity = identityProvider()
```
adelag-attentive, line 341: same thing as the other pr - put this identity check above "if !isPreviousIsRead" for better logic. all guard statements need to be up front

Sources/Inbox/InboxManager.swift
```diff
@@ -211,9 +210,35 @@ actor InboxManager {
...
+            return
+        }
+
+        do {
+            let response = try await api.markMessagesRead(
+                pushToken: identity.pushToken,
+                visitorId: identity.visitorId,
+                messageIds: [messageID]
+            )
+            applyReadStatusResponse(response)
```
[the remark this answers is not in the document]
usharif-attentive: Good catch — fixed in 105ecf8. `markRead` now snapshots `generation` before the PATCH and discards the response (on both the success and revert paths) if it advanced while the request was in flight, so a `resetUnreadCount()` (clearUser/updateUser) or a newer `refreshUnreadCount()` can't be clobbered by a stale mark-read response. Added `testMarkRead_identityChangeDuringRequest_discardsStaleResponse` to cover it.

Sources/API/ATTNAPI.swift
usharif-attentive: Done in 105ecf8. The mark-read request now sends `apns:<token>` via a small `inboxPushToken(_:)` helper (empty token still omits `push_token`). Note the unread-count and messages-list requests live on the base branch (MSDK-413) — want me to add the same prefix there, or are you handling those in that PR?

Sources/Inbox/InboxManager.swift
usharif-attentive: Done in 105ecf8 — moved the empty-visitor-id guard above the optimistic `isRead` flip so all guards are up front.

usharif-attentive:
Addressed review feedback in 105ecf8:

- **Stale response after identity change (P1):** `markRead` snapshots `generation` before the PATCH and drops the response on both the apply and revert paths if it advanced in flight. New test: `testMarkRead_identityChangeDuringRequest_discardsStaleResponse`.
- **Guard ordering:** empty-visitor-id guard moved above the optimistic flip.
- **`apns:` push-token prefix:** applied to the mark-read request via an `inboxPushToken(_:)` helper.
- **Codable request model (Codex non-blocking):** folded in at the base of the stack — replaced the `JSONSerialization` `[String: Any]` payload with a shared `UpdateReadStatusRequest: Encodable` (`encodeIfPresent` omits an empty `push_token`). The mark-unread follow-up (#274) can reuse the same model.

All 211 unit tests pass on the iOS 26.5 simulator (iPhone 17 Pro); lint clean (0 serious violations).

adelag-attentive:
overall looks good! but there are a bunch of (hopefully not too annoying) merge conflicts. maybe let's get my "get all messages" PR merged in since we're touching the same files anyway, then we address all the merge conflicts here in one go? otherwise, claude didn't flag anything else.

Sources/API/ATTNAPI.swift
```diff
@@ -490,6 +490,65 @@ final class ATTNAPI: ATTNAPIProtocol {
         }
     }
 
+    /// Marks the supplied messages as read on the server.
+    ///
+    /// PATCH /inbox/messages/read — identifier-based (visitor_id + push_token), body carries
+    /// the list of message ids. The response echoes the per-message read status and the
+    /// updated unread count so the caller can reconcile local state without a follow-up fetch.
+    func markMessagesRead(
```
adelag-attentive, line 498: **1. `markMessagesRead` bypasses the shared inbox HTTP helper — ~50 lines of duplicate scaffolding.**

Develop now has `postInboxJSON<T:>(path:payload:decoder:)` and `inboxIdentityPayload(...)` (used by `fetchInboxUnreadCount` / `fetchInboxMessages`). This func duplicates URL construction, headers, status check, and decode/error mapping — the behavior is identical, but the drift risk is real (e.g. the `"https://mobile.attentivemobile.com/inbox/messages/read"` string literal instead of `Self.inboxHost + "/inbox/messages/read"`).

One wrinkle: `postInboxJSON` hardcodes `POST` and this endpoint is `PATCH`. Cheapest fix is to add an `httpMethod: String = "POST"` parameter to `postInboxJSON` so mark-read and the upcoming mark-unread can share it. The func here should then reduce to ~15 lines: build the encoded body, call the helper.

Sources/API/ATTNAPI.swift
```diff
@@ -490,6 +490,65 @@ final class ATTNAPI: ATTNAPIProtocol {
...
+        } catch {
+            Loggers.network.error("Failed to decode inbox mark-read response: \(error.localizedDescription, privacy: .public)")
+            throw ATTNError.inboxResponseDecodeFailed
+        }
+    }
+
+    /// The inbox backend expects push tokens namespaced by transport (e.g. `apns:<token>`).
+    /// Returns `nil` for an empty token so `push_token` is omitted from the request body
+    /// rather than sent as an empty (and unusable) identifier.
+    private static func inboxPushToken(_ pushToken: String) -> String? {
```
adelag-attentive, line 548: **2. `inboxPushToken` re-implements what `inboxIdentityPayload` already does.**

`inboxIdentityPayload` (currently at ~line 432 on develop) already omits `push_token` when empty and applies the `apns:` prefix. Two implementations of the same rule will drift. If mark-read's contract (no `c`/email/phone) means you can't reuse `inboxIdentityPayload` wholesale, at minimum let this static helper live next to it and have `inboxIdentityPayload` call it, so the two can't get out of sync.

Sources/API/ATTNAPI.swift
```diff
@@ -490,6 +490,65 @@ final class ATTNAPI: ATTNAPIProtocol {
...
+            throw ATTNError.inboxResponseDecodeFailed
+        }
+        Loggers.network.debug("Inbox mark-read status code: \(http.statusCode, privacy: .public)")
+        guard (200..<300).contains(http.statusCode) else {
+            Loggers.network.error("Inbox mark-read API returned status \(http.statusCode, privacy: .public)")
+            throw ATTNError.inboxRequestFailed(statusCode: http.statusCode)
+        }
+
+        do {
+            return try JSONDecoder().decode(UpdateReadStatusResponse.self, from: data)
```
adelag-attentive, line 538: **3. Uses bare `JSONDecoder()` instead of `Self.inboxJSONDecoder`.**

Every other inbox decode path uses `Self.inboxJSONDecoder`, which tolerates ISO-8601 with or without fractional seconds. The response body has no dates today so this doesn't manifest — but the moment the server adds a `read_at` timestamp to the mark-read/mark-unread response, every mark call will fail with `.inboxResponseDecodeFailed`. Trivial swap now, prevents a landmine later.

Sources/Inbox/InboxManager.swift
```diff
@@ -332,10 +333,44 @@ actor InboxManager {
...
+                pushToken: identity.pushToken,
+                visitorId: identity.visitorId,
+                messageIds: [messageID]
+            )
+            guard generation == refreshGeneration else { return }
+            applyReadStatusResponse(response)
+        } catch {
+            Loggers.network.error("Failed to mark inbox message read: \(error.localizedDescription, privacy: .public)")
+            guard generation == refreshGeneration else { return }
+            if messagesByID[messageID] != nil, !previousIsRead {
```
adelag-attentive, line 369: **4. Failure revert races another mark-read on the same id.**

User taps mark-read twice quickly on the same message:
- Call A: `previousIsRead = false`, flips local to `true`, PATCH_A in flight.
- Call B: `previousIsRead = true`, skips local flip, PATCH_B in flight.
- PATCH_A fails → catch branch sees `!previousIsRead` (true) → reverts to `false`.
- PATCH_B succeeds → `applyReadStatusResponse` sets `isRead = true`.

Order of arrival now decides final state; if PATCH_B lands *before* PATCH_A's failure, the revert clobbers a legitimate server-confirmed `true`. The `refreshGeneration` guard doesn't help — neither call bumped it.

Two options:
1. Track pending mark-read IDs in a `Set<Message.ID>`; skip the revert if another op for the same id is still in flight.
2. Drop the revert branch entirely and let a subsequent `applyReadStatusResponse` or the next `refresh()` be the source of truth (the optimistic flip is already surfaced to the UI; leaving it in place on failure isn't obviously wrong).

adelag-attentive (commented):
Reviewed the mark-read wiring. Overall design is sound (optimistic flip + server-authoritative reconciliation + revert-on-failure), and test coverage is thorough. Four critical items below — #1 is the biggest lift and drives most of the diff.

adelag-attentive:
### Follow-up: mark-read chattiness under the RFC's viewport trigger

Circling back on the "already-read messages now still fire a PATCH" behavior change. Cross-referenced against the [RFC (In-App Inbox Messaging)](https://docs.google.com/document/d/1vlzpxT7vR51kH5dotCmN1e6D09XLVvj2T9MRtCCUiVc/edit) — the mark-read section explicitly states:

> On the SDK side, this endpoint will be called when ≥75% of the message cell is visible for at least 1 continuous second.

That's viewport-based, not tap-based. In practice, scrolling through a 50-message inbox where 45 are already read triggers **50 PATCHes** under the current code vs. **5** under the pre-#273 behavior. At JCP-scale traffic (~618K daily app launches, ~797K peak per the RFC) that's a meaningful volume amplification against `/inbox/messages/read` for zero product value on the redundant calls, and it amplifies the current mocked-backend bug where every response hardcodes `unread_count = 5`.

Not blocking this PR — the correctness is fine and the test coverage is solid. Filing a follow-up ticket to address in a subsequent MSDK release. Two options for the fix, non-exclusive:

**Option 1 — Local-cache guard (minimal, matches pre-#273 behavior).** Skip the PATCH entirely if the message is already `isRead == true` locally. Restores the old short-circuit:
```swift
func markRead(_ messageID: Message.ID) async {
    guard let previousIsRead = messagesByID[messageID]?.isRead else { return }
    guard !previousIsRead else { return } // <-- restored
    // ...existing optimistic flip + PATCH + reconcile...
}
```
Tradeoff: loses cross-device reconciliation on redundant marks. Not a real capability loss — cross-device sync is already handled by `refreshMessages` / `refreshUnreadCount` on foreground and push open.

**Option 2 — Batch under the visibility emitter.** The RFC endpoint already accepts a plural `message_ids` array, and the current SDK sends one message at a time (`messageIds: [messageID]`) — leaving batching on the table regardless. Debounce the visibility tracker so a scrolling session emits **one** PATCH with N ids instead of N PATCHes. Rough shape:
```swift
// In the visibility-tracker layer:
private var pendingReadIds: Set<Message.ID> = []
private var flushTask: Task<Void, Never>?

func onMessageViewed(_ id: Message.ID) {
    pendingReadIds.insert(id)
    flushTask?.cancel()
    flushTask = Task { [weak self] in
        try? await Task.sleep(for: .milliseconds(500))
        await self?.flushPendingReads()
    }
}
```
Then `InboxManager.markRead` accepts `[Message.ID]` and forwards the batch to `api.markMessagesRead(messageIds:)` in one call.

**Recommendation:** ship both. Option 1 keeps single explicit calls (swipe-to-mark-read) efficient; Option 2 keeps viewport-triggered calls efficient. A follow-up ticket will be created under [MSDK-198](https://attentivemobile.atlassian.net/browse/MSDK-198) to track this work.

[MSDK-198]: https://attentivemobile.atlassian.net/browse/MSDK-198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

adelag-attentive (approved):
overall good enough to ship - let's merge this to unblock the other PR

PR #274, opened 2026-07-09
MSDK-411: wire real endpoint for mark message unread

usharif-attentive:
## Summary

Wires `markUnread` to the real `PATCH /inbox/messages/unread` endpoint, replacing the previous local-only mutation. Part of the inbox real-endpoint rollout ([MSDK-411](https://attentivemobile.atlassian.net/browse/MSDK-411)).

**Stacked on top of [MSDK-410](https://attentivemobile.atlassian.net/browse/MSDK-410) — this PR targets `umair/MSDK-410-real-mark-read-endpoint`. Review/merge 410 first.** The diff here is only the unread-specific additions; the shared `UpdateReadStatusResponse` model and mark scaffolding land in 410.

## Key Changes

- `ATTNAPI.markMessagesUnread(pushToken:visitorId:messageIds:)` issues `PATCH /inbox/messages/unread` with the identifier-based body `{ visitor_id, push_token, message_ids }` and decodes `UpdateReadStatusResponse`.
- `InboxManager.markUnread` is now `async`: it optimistically flips `isRead = false` locally, then reconciles per-message read status and the server-authoritative `unread_count` from the response. On failure it reverts the local flip.

## Public API Impact

`ATTNSDK.markUnread(for:)` was already `public` and `async`; signature unchanged, all call sites already `await`. No source/binary break. Behavior changes from local-only to a network call.

## Verification

- All 211 unit tests pass via `xcodebuild test` on the iOS 26.5 simulator (iPhone 17 Pro). Adds 6 API-level tests (`ATTNInboxMarkUnreadAPITests`) and 4 `InboxManagerTests` cases (success reconciles `unread_count`, failure reverts, already-unread still calls the API, unknown id no-ops), mirroring the mark-read coverage.
- `bundle exec fastlane ios lint` passes (0 serious violations).
- Exercised end-to-end in Bonni: swiping mark-unread fires `PATCH /inbox/messages/unread`, returns 200, and the SDK correctly decodes/stores/emits. **Note:** the prod backend currently routes to `MockInboxMessageClient`, which hardcodes `unread_count = 5` in every response, so the badge count does not move yet. The per-message flip is echoed correctly by the server. Badge movement will work once the real `InboxMessageClient` is swapped in server-side (raised with the backend team).

## Risk

Medium. Contained to the inbox mark-unread path. Optimistic-flip + revert keeps the UI truthful on network failure. No change to identity, events, or push handling. Recent backend flakiness on inbox endpoints (intermittent 401s in prod) applies here too.

## Observability

`Loggers.network` OSLog entries (`PATCH /inbox/messages/unread`, status code, `Failed to mark inbox message unread`) on-device; `mobile-api` service logs (`Received mark messages unread request`) server-side. No dedicated dashboard for this endpoint yet.

## Rollback

No feature flag or config kill-switch — mitigation is a follow-up SDK release plus host-app re-adoption. Client-side revert-on-failure limits blast radius.

[MSDK-411]: https://attentivemobile.atlassian.net/browse/MSDK-411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MSDK-410]: https://attentivemobile.atlassian.net/browse/MSDK-410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Inbox/InboxManager.swift
+++ b/Sources/Inbox/InboxManager.swift
@@ -52,10 +52,10 @@ actor InboxManager {
     /// load returning inside the refresh window resets that flag but must not release this one.
     private var isRefreshingFirstPage: Bool = false
 
-    /// Server-authoritative unread count. Written by `refreshUnreadCount()` and by the mark-read
-    /// API response (see `applyReadStatusResponse`); local `markUnread` does not touch it. Reads
-    /// should go through the `unreadCount` accessor, which awaits the init-time refresh so the
-    /// first read never returns a stale 0.
+    /// Server-authoritative unread count. Written by `refreshUnreadCount()` and by the mark-read /
+    /// mark-unread API responses (see `applyReadStatusResponse`). Reads should go through the
+    /// `unreadCount` accessor, which awaits the init-time refresh so the first read never returns
+    /// a stale 0.
     private var storedUnreadCount: Int = 0
 
     private let api: ATTNAPIProtocol
@@ -373,10 +373,44 @@ actor InboxManager {
         }
     }
 
-    func markUnread(_ messageID: Message.ID) {
-        guard messagesByID[messageID]?.isRead == true else { return }
-        messagesByID[messageID]?.isRead = false
-        send(.loaded(orderedMessagesSnapshot()))
+    /// Marks a message unread. Optimistically flips the local state, then reconciles from the
+    /// server response. On failure the local flip is reverted so the UI reflects true state.
+    func markUnread(_ messageID: Message.ID) async {
+        guard let previousIsRead = messagesByID[messageID]?.isRead else { return }
+
+        let identity = identityProvider()
+        guard !identity.visitorId.isEmpty else {
+            Loggers.network.debug("Skipping inbox mark-unread: empty visitor ID")
+            return
+        }
+
+        if previousIsRead {
+            messagesByID[messageID]?.isRead = false
+            send(.loaded(orderedMessagesSnapshot()))
+        }
+
+        // Snapshot the refresh generation before the request. If an identity change
+        // (`resetForIdentityChange`) or a newer `refreshUnreadCount` lands while the PATCH is in
+        // flight, the generation advances and we drop this now-stale response instead of
+        // restoring an old unread count.
+        let generation = refreshGeneration
+
+        do {
+            let response = try await api.markMessagesUnread(
+                pushToken: identity.pushToken,
+                visitorId: identity.visitorId,
+                messageIds: [messageID]
+            )
+            guard generation == refreshGeneration else { return }
+            applyReadStatusResponse(response)
+        } catch {
+            Loggers.network.error("Failed to mark inbox message unread: \(error.localizedDescription, privacy: .public)")
+            guard generation == refreshGeneration else { return }
+            if messagesByID[messageID] != nil, previousIsRead {
+                messagesByID[messageID]?.isRead = previousIsRead
+                send(.loaded(orderedMessagesSnapshot()))
+            }
+        }
     }
 
     func delete(_ messageID: Message.ID) {
--- a/Tests/TestCases/ATTNInboxMarkUnreadAPITests.swift
+++ b/Tests/TestCases/ATTNInboxMarkUnreadAPITests.swift
@@ -1,0 +1,132 @@
+//
+//  ATTNInboxMarkUnreadAPITests.swift
+//  attentive-ios-sdk Tests
+//
+
+import XCTest
+@testable import ATTNSDKFramework
+
+final class ATTNInboxMarkUnreadAPITests: XCTestCase {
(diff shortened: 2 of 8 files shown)


usharif-attentive:
@codex review

usharif-attentive:
@attentive-claude review

Sources/Inbox/InboxManager.swift
```diff
@@ -241,9 +241,35 @@ actor InboxManager {
...
+    func markUnread(_ messageID: Message.ID) async {
+        guard let previousIsRead = messagesByID[messageID]?.isRead else { return }
+        if previousIsRead {
+            messagesByID[messageID]?.isRead = false
+            send(.loaded(allMessages))
+        }
+
+        let identity = identityProvider()
+        guard !identity.visitorId.isEmpty else {
+            Loggers.network.debug("Skipping inbox mark-unread: empty visitor ID")
```
adelag-attentive, line 383: should we move this to before the if previousIsRead, to prevent it doing unnecessary work?

Tests/TestCases/ATTNInboxMarkUnreadAPITests.swift
```diff
@@ -0,0 +1,131 @@
...
+
+    override func tearDown() {
+        sessionMock = nil
+        api = nil
+        userIdentity = nil
+        super.tearDown()
+    }
+
+    private func markUnread(
+        pushToken: String = "fcm:abc123",
```
adelag-attentive: nit: we should probably use apns here

Sources/Inbox/InboxManager.swift
```diff
@@ -241,9 +241,35 @@ actor InboxManager {
...
+            return
+        }
+
+        do {
+            let response = try await api.markMessagesUnread(
+                pushToken: identity.pushToken,
+                visitorId: identity.visitorId,
+                messageIds: [messageID]
+            )
+            applyReadStatusResponse(response)
```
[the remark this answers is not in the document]
usharif-attentive: Fixed in 41ab0bb. `markUnread` now snapshots `refreshGeneration` before the PATCH and discards the response (on both the apply and revert paths) if it advanced while in flight — so an identity reset (`resetForIdentityChange`, formerly the `generation` bump) or a newer `refreshUnreadCount` can't be clobbered by a stale mark-unread response. Added `testMarkUnread_identityChangeDuringRequest_discardsStaleResponse` to cover it. Mirrors the same fix in #273.

Sources/Inbox/InboxManager.swift
usharif-attentive: Done in 41ab0bb — moved the empty-visitor-id guard above the optimistic flip so all guards are up front and we don't do the local work (or emit a state change) when we're going to bail anyway.

Tests/TestCases/ATTNInboxMarkUnreadAPITests.swift
usharif-attentive: Done in 41ab0bb. `markMessagesUnread` now sends `apns:<token>` via the shared `inboxPushToken(_:)` helper introduced in #273, and this test asserts the `apns:` prefix. The request body also now uses the shared `UpdateReadStatusRequest` Codable model rather than manual `JSONSerialization`, matching mark-read.

usharif-attentive:
Rebased onto the updated MSDK-410 branch (which now sits on `develop` after #275/#266 merged) and addressed review feedback in `41ab0bb`:

- **Stale response after identity change (P2):** `markUnread` snapshots `refreshGeneration` before the PATCH and drops the response on both the apply and revert paths if it advanced in flight. New test: `testMarkUnread_identityChangeDuringRequest_discardsStaleResponse`.
- **Guard ordering:** empty-visitor-id guard moved above the optimistic flip.
- **`apns:` push-token prefix:** `markMessagesUnread` uses the shared `inboxPushToken(_:)` helper; the API test asserts the prefix.
- **Codable request model:** reuses the `UpdateReadStatusRequest` model from #273 instead of manual `JSONSerialization`.

Also adapted `markUnread`/`applyReadStatusResponse` to develop's evolved inbox model (`storedUnreadCount` + async `unreadCount` accessor, `orderedMessagesSnapshot()`, `resetForIdentityChange()`).

Full suite passes on the iOS 26.5 simulator (iPhone 17 Pro). Still stacked on #273 — merge that first.

adelag-attentive:
The comment mentioned merging https://github.com/attentive-mobile/attentive-ios-sdk/pull/273 first, would you like to target resolving that? @usharif-attentive

adelag-attentive (approved):

PR #275, opened 2026-07-16
MSDK-413: wire real endpoint for inbox message list

adelag-attentive:
## Summary

Wires the SDK's inbox to the real `POST /inbox/messages` endpoint, replacing the in-memory mock generator. Adds pull-up-to-load-more infinite scroll with a footer spinner and dedup across pages.

- **Endpoint:** `ATTNAPI.fetchInboxMessages` mirrors `fetchInboxUnreadCount` (same host, `c` domain + visitor_id + `apns:`-prefixed push_token + email + phone). Payload also carries `page_size` and optional `page_token`. Both endpoints now share a `postInboxJSON<T:>` + `inboxIdentityPayload()` scaffolding.
- **Manager (`InboxManager`):** first-page fetch on init, `loadNextPage()` for infinite scroll (returns `Bool` so the ViewModel can gate the spinner without a probe hop), dedup on message id, generation counters that discard stale-identity responses, `resetForIdentityChange()` clears messages+cursor+count on `clearUser`/`updateUser`. Init + `refresh()` fire messages and unread-count fetches concurrently via `async let`.
- **UI (`InboxView` / `InboxViewModel`):** `.onAppear` on the last row triggers `loadNextPage()`; footer `ProgressView` shows while `isLoadingMore == true`.
- **Message model:** decodes `inbox_message_id`, `sent_at`, `expires_at`, `read_at`, `is_read`, `image_url`, `action_url`; `style` defaults to `.small` when absent so encode-decode round-trips preserve the field. Inbox JSON decoder tolerates ISO-8601 with or without fractional seconds.

## Test plan

- [x] Unit tests: 24 inbox tests pass locally (API payload assembly, error mapping, decode failures, pagination dedup, identity reset clears messages, refresh error preserves cursor, `apns:` prefix asserted).
- [ ] Manual: open Bonni inbox, verify first page loads, scroll to bottom to trigger next page (footer spinner appears), pull to refresh, mark read/unread, `clearUser` from settings and confirm inbox empties before next identity's messages load.
- [ ] CI: SwiftLint + framework build green.
simulator_screenshot_456F5B13-05F2-48C9-817A-65FDE5D95356

### DIFF
--- a/Sources/Inbox/InboxManager.swift
+++ b/Sources/Inbox/InboxManager.swift
@@ -66,7 +66,7 @@ actor InboxManager {
     var allMessages: [Message] {
         get async {
             await awaitInitialRefresh()
-            return messageOrder.compactMap { messagesByID[$0] }
+            return orderedMessagesSnapshot()
         }
     }
 
@@ -77,8 +77,8 @@ actor InboxManager {
         }
     }
 
-    /// True when the last successful fetch reported a `next_page_token`. Used by the UI to
-    /// decide whether to attempt a pull-up load.
+    /// True when the last successful fetch reported a `next_page_token`. Used by tests and any
+    /// caller that wants a cheap "can we page further" signal without calling `loadNextPage()`.
     var hasMore: Bool {
         nextPageToken?.isEmpty == false
     }
@@ -121,24 +121,28 @@ actor InboxManager {
         // `ATTNSDK.materializedInboxManager()`. Calls `performUnreadCountFetch` directly (not
         // `refreshUnreadCount`) so it doesn't try to coalesce with itself.
         initialRefreshTask = Task { [weak self] in
-            await self?.loadInitialMessagesAndUnreadCount()
+            await self?.performMessagesAndUnreadCountRefresh()
         }
     }
 
-    /// Fetches the first page of inbox messages, replacing any previously loaded messages.
-    /// Callers should invoke this on inbox open / pull-to-refresh / push open.
+    /// Fetches the first page of inbox messages and the latest unread count, replacing any
+    /// previously loaded messages. The two fetches run concurrently. Callers should invoke
+    /// this on inbox open / pull-to-refresh / push open.
     func refresh() async {
-        await performMessagesRefresh()
+        await performMessagesAndUnreadCountRefresh()
     }
 
     /// Fetches the next page of messages using the stored `nextPageToken` and appends them.
     /// Safe to call repeatedly — no-ops when no more pages are available or another fetch is in flight.
-    func loadNextPage() async {
+    /// Returns `true` when a network fetch was actually started; callers can use this to gate a
+    /// "loading more" indicator without needing a separate probe.
+    @discardableResult
+    func loadNextPage() async -> Bool {
         guard hasFetchedFirstPage, !isLoadingNextPage, let pageToken = nextPageToken, !pageToken.isEmpty else {
-            return
+            return false
         }
         let identity = identityProvider()
-        guard !identity.visitorId.isEmpty else { return }
+        guard !identity.visitorId.isEmpty else { return false }
 
         isLoadingNextPage = true
         let generation = messagesGeneration
@@ -155,7 +159,7 @@ actor InboxManager {
             guard generation == messagesGeneration else {
                 // Identity or refresh() moved on; discard this page.
                 isLoadingNextPage = false
-                return
+                return true
             }
             appendMessages(response.messages)
             nextPageToken = response.nextPageToken
@@ -166,6 +170,7 @@ actor InboxManager {
             isLoadingNextPage = false
             // Keep existing list; caller can retry by pulling up again.
         }
+        return true
     }
 
     /// Fetches the latest unread count from the server and stores it as the
@@ -183,12 +188,14 @@ actor InboxManager {
         await performUnreadCountFetch(skipNotify: false)
     }
 
-    /// Init-time seed: fetch the first page of messages and the server-authoritative unread
-    /// count. The count fetch runs with `skipNotify: false` so subscribers get a nudge to
-    /// re-read `unreadCount` when the server value arrives.
-    private func loadInitialMessagesAndUnreadCount() async {
-        await performMessagesRefresh()
-        await performUnreadCountFetch(skipNotify: false)
+    /// Fetches the first page of messages and the server-authoritative unread count concurrently.
+    /// Shared by init-time load and every `refresh()`. `skipNotify: false` lets subscribers get a
+    /// nudge to re-read `unreadCount` when the count arrives after the messages have already
+    /// emitted `.loaded` (order between the two `async let`s is non-deterministic).
+    private func performMessagesAndUnreadCountRefresh() async {
+        async let messages: Void = performMessagesRefresh()
+        async let count: Void = performUnreadCountFetch(skipNotify: false)
+        _ = await (messages, count)
     }
 
     /// Internal worker used by both the init task and the public `refresh()`. Fetches page 1,
@@ -202,9 +209,7 @@ actor InboxManager {
 
         messagesGeneration &+= 1
         let generation = messagesGeneration
-        // Reset pagination bookkeeping so a stale nextPageToken can't accidentally leak into
-        // this refresh's follow-up load.
-        nextPageToken = nil
+        // Cancel any in-flight page load; the bumped generation will make it a no-op on return.
         isLoadingNextPage = false
 
         // Only emit .loading on the very first refresh; on subsequent refreshes keep the last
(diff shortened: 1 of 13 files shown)


Sources/Inbox/InboxManager.swift
```diff
@@ -126,19 +196,66 @@ actor InboxManager {
...
+        let identity = identityProvider()
+        guard !identity.visitorId.isEmpty else {
+            Loggers.network.debug("Skipping inbox messages refresh: empty visitor ID")
+            return
+        }
+
+        messagesGeneration &+= 1
+        let generation = messagesGeneration
+        // Cancel any in-flight page load; the bumped generation will make it a no-op on return.
+        isLoadingNextPage = false
```
usharif-attentive: **Concurrency gap: a `loadNextPage()` that starts _during_ this refresh isn't discarded.**

Setting `isLoadingNextPage = false` here (after bumping `messagesGeneration`) correctly no-ops any page load that started *before* the bump — that one captured the old generation. But it also *opens the gate* for a new `loadNextPage()` that fires during the page-1 await window (e.g. last-row `.onAppear` during a pull-to-refresh): it passes the guard (`isLoadingNextPage == false`, `nextPageToken` still the pre-refresh cursor) and captures the *same* generation as this refresh.

When both complete, neither is discarded:
- refresh → `replaceMessages(page1)` + sets the fresh `nextPageToken`
- then loadNextPage → `appendMessages(stalePage)` + **overwrites `nextPageToken` with the stale page's token**

So a stale page gets merged into the refreshed list and the pagination cursor is clobbered (dedup masks overlapping ids, but not the token clobber or non-overlapping stale rows). It self-heals on the next refresh, but the generation mechanism is meant to prevent exactly this.

Suggest blocking page loads for the duration of the refresh instead of opening the gate — e.g. keep `isLoadingNextPage = true` while the refresh is in flight, or gate `loadNextPage` on a "refresh in progress" flag.

Sources/Inbox/InboxViewModel.swift
```diff
@@ -45,6 +49,22 @@ class InboxViewModel: ObservableObject {
...
+    /// its `Bool` return tells us whether a fetch actually started so we can toggle the footer
+    /// spinner off only when there was one to hide.
+    func loadNextPage() {
+        // Optimistically show the spinner; the manager will either fetch (spinner stays until the
+        // fetch returns) or no-op (we clear it immediately below). A brief flash on no-op is
+        // preferable to two actor hops on every last-row `.onAppear`.
+        isLoadingMore = true
+        Task { [weak self] in
+            guard let self = self else { return }
+            _ = await self.inboxManager.loadNextPage()
```
usharif-attentive: **The `Bool` return is discarded, which contradicts the stated design.**

The PR description and `InboxManager.loadNextPage()`'s doc comment sell this `Bool` as letting the ViewModel "gate the spinner without a probe hop," but here it's `_ = await …` and `isLoadingMore` is set to `false` unconditionally. Two consequences:

1. `isLoadingMore = true` is set optimistically on *every* last-row `.onAppear`, even when there's no next page → brief spinner flash on scroll-to-bottom (the mechanism to avoid this was the return value, now unused).
2. Rapid `.onAppear` events spawn multiple Tasks; a no-op Task returns immediately and sets `isLoadingMore = false` while the real fetch Task is still running → the spinner can vanish mid-fetch.

Either use the return value to drive the spinner (only show it when the manager reports a fetch actually started), or drop the `Bool` from the manager API and simplify the docs so code and intent agree.

Sources/Inbox/InboxManager.swift
adelag-attentive: Addressed in c6b8c8c — added an isRefreshingFirstPage flag with a defer to guarantee release, and loadNextPage now refuses while it's set. Test testLoadNextPage_firedDuringRefresh_isBlocked reproduces the race via a new onFetchInboxMessages hook on ATTNAPISpy and asserts the fresh cursor survives.

Sources/Inbox/InboxViewModel.swift
adelag-attentive: Addressed in c6b8c8c — dropped the Bool return from loadNextPage entirely. Added a loadingMoreStream on the manager that fans out every isLoadingNextPage transition (via a single setLoadingNextPage setter). The ViewModel subscribes to it, so isLoadingMore always reflects the manager's true fetch state: no flash on no-op .onAppear, and a fast no-op Task can't clear it while a real fetch is still running. Test testLoadingMoreStream_emitsTransitionsForRealFetchesOnly covers it.

Sources/Inbox/InboxManager.swift
usharif-attentive: Thanks — the gate closes the flagged window, but two follow-ups before I can approve:

**1. The gate leaks under overlapping refreshes.** `isRefreshingFirstPage` is a Bool released by an unconditional `defer`. If two `refresh()` calls overlap (double pull-to-refresh, `.task` racing `.refreshable`, or `refreshUnreadCount()` nil-ing `initialRefreshTask` so a concurrent `refresh()` bypasses coalescing), the superseded refresh fails its generation check and early-returns — and its `defer` clears the flag **while the newer refresh is still in flight**. A `loadNextPage()` in that sub-window passes every guard, captures the live generation, and lands after the newer refresh settles: stale rows appended + `nextPageToken` clobbered — the original race one reentrancy level deeper.

Same shape in `loadNextPage`'s own `defer setLoadingNextPage(false)`: a slow, generation-discarded page load can clear a *newer* page load's flag → spinner hides mid-fetch and a duplicate same-page fetch slips past `!isLoadingNextPage`.

Two-line fix — guard both defers on generation (supersession is safe because whoever bumps the generation — a newer refresh or `resetForIdentityChange` — already clears both flags itself):
```swift
defer { if generation == messagesGeneration { isRefreshingFirstPage = false } }   // performMessagesRefresh
defer { if generation == messagesGeneration { setLoadingNextPage(false) } }       // loadNextPage
```
(or make the gate a counter.)

**2. `testLoadNextPage_firedDuringRefresh_isBlocked` passes even without the fix**, so it doesn't guard the regression: the hook awaits `loadNextPage()` inline inside the spy's `fetchInboxMessages`, so on unfixed code the stale append/clobber fully settles *before* the refresh's response returns — the refresh's generation still matches and `replaceMessages` wipes the stale state, satisfying both assertions. Adding `XCTAssertEqual(apiSpy.fetchInboxMessagesCallCount, 2)` makes it discriminate (unfixed code fires 3).

Sources/Inbox/InboxViewModel.swift
usharif-attentive: The stream mechanism is sound — no-op `.onAppear` calls can't flash the spinner anymore, subscription lifecycle / MainActor delivery / continuation cleanup all check out. Two residual items:

**1. The premature-spinner-clear symptom isn't fully gone.** `loadNextPage`'s unconditional `defer setLoadingNextPage(false)` runs even when the response was discarded as stale. Interleaving: page load A in flight → `refresh()` bumps the generation and clears the flag → refresh completes → page load B starts (flag true, spinner shows) → A's slow response returns, fails the generation check, and its `defer` clears **B's** flag — `loadingMoreStream` emits `false` while B is genuinely fetching, and a further last-row `.onAppear` now passes `!isLoadingNextPage` and duplicate-fetches B's page token. Generation-guarding the defer (see the InboxManager thread) fixes this too.

**2. `testLoadingMoreStream_emitsTransitionsForRealFetchesOnly` can't catch the regression it's named for.** The collector breaks on `!next && observed.count >= 3`, i.e. right after the *first* (real) fetch's `[false, true, false]` — so if the no-op second call wrongly emitted `true→false`, nothing observes it and the assertion still passes. And if `setLoadingNextPage` regressed to under-emit, the collector blocks forever on `iter.next()` and the test hangs instead of failing. Suggest: keep collecting until after both calls complete (cancel the collector to end the stream), assert the full emission list, and add a `fetchInboxMessagesCallCount` assertion so the no-op half is checked directly.

Sources/Inbox/InboxManager.swift
adelag-attentive: Addressed in ea6fff2. Both defers now guard release on `generation == messagesGeneration`, so a superseded refresh or page load can't clear the newer one's flag. Note on the counter alternative: since `resetForIdentityChange` bumps `messagesGeneration` and clears both flags itself, and a newer `performMessagesRefresh` bumps + re-sets the flags, guarding the defer is sufficient without switching to a counter. Also strengthened the test: added `XCTAssertEqual(apiSpy.fetchInboxMessagesCallCount - callCountBeforeRefresh, 1)` so the racing loadNextPage being blocked is checked directly — unfixed code would see +2, fixed sees +1.

Sources/Inbox/InboxViewModel.swift
adelag-attentive: Addressed in ea6fff2. `loadNextPage`'s defer now guards release on generation, so a stale response returning after a newer page load can't clear the newer one's flag. Also strengthened the test: the collector now runs for the full duration of both calls and is cancelled by the test after both settle, then asserts the complete emission list — a wrongly-emitted `true→false` from the no-op call would now be caught. Added an `XCTAssertEqual(apiSpy.fetchInboxMessagesCallCount - callCountBeforePaging, 1)` assertion so the no-op half is verified directly.

usharif-attentive (approved):

PR #276, opened 2026-07-20
MSDK-412: wire real endpoint for inbox click tracking

adelag-attentive:
## Summary

Wires inbox message clicks to `POST /inbox/events/clicked` per the RFC ([MSDK-412](https://attentivemobile.atlassian.net/browse/MSDK-412)).

## Key changes

- **`ATTNAPI.markMessageClicked(pushToken:visitorId:messageId:actionURL:)`** — POSTs the RFC's minimal identifier-based body (`visitor_id` + `apns:`-prefixed `push_token` + `message_id` + `action_url`). No `c`, no email/phone — matches the RFC contract exactly. Expects 204 No Content on success.
- **Helper refactor** — `postInboxJSON<T:>`'s request scaffolding split into a private `postInboxRaw` worker. New sibling `postInbox(path:payload:)` returns `Void` for endpoints without a response body. Existing decode-based callers unchanged.
- **`InboxManager.markClicked(_:)`** — flips local `isRead` (matching iOS Mail UX) and fires the API call fire-and-forget. Failures are logged via `Loggers.network`, never surfaced to the UI (nothing to reconcile from a 204).
- **UI** — `InboxView` row now has a tap gesture that fires `viewModel.click(id)`. The SDK deliberately does not open `actionURL` — host apps handle deep-link routing.
- **Public API** — `sdk.markClicked(for: messageID)` mirrors `markRead`/`markUnread`/`delete` for hosts rendering their own inbox UI.

## Public API impact

New `public func markClicked(for:) async` on `ATTNSDK`. Additive only; no source or binary breaks.

## Verification

- 42 inbox tests pass locally (`xcodebuild test`, iPhone 16 simulator).
  - 9 API-level: URL, method, body shape, `apns:` prefix, empty push token, nil/empty actionURL, server error, network error, 204 no-body, 200-with-body.
  - 6 manager-level: fires with expected args, flips local read state, already-read still tracked, unknown id no-ops, empty visitor skips network, API failure is swallowed.
- Framework build ✅.

## Test plan

- [ ] Manual: open Bonni inbox, tap a message, verify `POST /inbox/events/clicked` fires with `apns:` prefix and correct `action_url`. Row's unread dot clears.
- [ ] Tap on an already-read message still fires the click POST.
- [ ] Network failure (airplane mode) doesn't crash or surface an error to the UI.
- [ ] `sdk.markClicked(for:)` fires the same endpoint when called from a host app's custom row.

## Risk

Low. Contained to the click path. No changes to identity, events, or push handling. Failure mode is silent (matches the analytics-fire-and-forget contract in the RFC).

[MSDK-412]: https://attentivemobile.atlassian.net/browse/MSDK-412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/API/ATTNAPI.swift
+++ b/Sources/API/ATTNAPI.swift
@@ -435,12 +435,10 @@ final class ATTNAPI: ATTNAPIProtocol {
         actionURL: String?
     ) async throws {
         Loggers.network.debug("Reporting inbox click - Visitor ID: \(visitorId, privacy: .public), Message ID: \(messageId, privacy: .public)")
-        var payload: [String: Any] = [
-            "c": domain,
-            "visitor_id": visitorId,
-            "message_id": messageId
-        ]
-        if !pushToken.isEmpty { payload["push_token"] = "apns:\(pushToken)" }
+        // Reuse `inboxIdentityPayload` so the `apns:` prefix + empty-token omission live in one place.
+        // Click contract carries no email/phone.
+        var payload = inboxIdentityPayload(pushToken: pushToken, email: nil, phone: nil, visitorId: visitorId)
+        payload["message_id"] = messageId
         if let actionURL = actionURL, !actionURL.isEmpty {
             payload["action_url"] = actionURL
         }
--- a/Sources/API/ATTNAPI.swift
+++ b/Sources/API/ATTNAPI.swift
@@ -426,6 +426,27 @@ final class ATTNAPI: ATTNAPIProtocol {
         return decoded
     }
 
+    /// Reports a message click to the server. Per RFC the endpoint carries only
+    /// `visitor_id` + `push_token` + `message_id` + `action_url` — no `c`, no email/phone —
+    /// and returns 204 No Content on success.
+    func markMessageClicked(
+        pushToken: String,
+        visitorId: String,
+        messageId: String,
+        actionURL: String?
+    ) async throws {
+        Loggers.network.debug("Reporting inbox click - Visitor ID: \(visitorId, privacy: .public), Message ID: \(messageId, privacy: .public)")
+        var payload: [String: Any] = [
+            "visitor_id": visitorId,
+            "message_id": messageId
+        ]
+        if !pushToken.isEmpty { payload["push_token"] = "apns:\(pushToken)" }
+        if let actionURL = actionURL, !actionURL.isEmpty {
+            payload["action_url"] = actionURL
+        }
+        try await postInbox(path: "/inbox/events/clicked", payload: payload)
+    }
+
     /// Builds the identifier fields shared by every inbox endpoint: `c` (domain), `visitor_id`,
     /// and the optional `push_token` / `email` / `phone`. Push token is prefixed with `apns:`
     /// so the server can route it to APNs (Android sends `fcm:...`).
@@ -449,14 +470,33 @@ final class ATTNAPI: ATTNAPIProtocol {
         return payload
     }
 
-    /// Shared POST + JSON-decode scaffolding for inbox endpoints. Maps non-2xx status to
-    /// `inboxRequestFailed`, non-HTTP responses and decode failures to `inboxResponseDecodeFailed`,
-    /// and bad URLs to `badURL`. All other transport errors propagate.
+    /// Shared POST + JSON-decode scaffolding for inbox endpoints that return a decodable body.
+    /// Maps non-2xx status to `inboxRequestFailed`, non-HTTP responses and decode failures to
+    /// `inboxResponseDecodeFailed`, and bad URLs to `badURL`. All other transport errors propagate.
     private func postInboxJSON<T: Decodable>(
         path: String,
         payload: [String: Any],
         decoder: JSONDecoder
     ) async throws -> T {
+        let data = try await postInboxRaw(path: path, payload: payload)
+        do {
+            return try decoder.decode(T.self, from: data)
+        } catch {
+            Loggers.network.error("Failed to decode inbox \(path, privacy: .public) response: \(error.localizedDescription, privacy: .public)")
+            throw ATTNError.inboxResponseDecodeFailed
+        }
+    }
+
+    /// Shared POST scaffolding for inbox endpoints that return no body (e.g. 204 No Content).
+    /// Same error mapping as `postInboxJSON`; response body (if any) is discarded.
+    private func postInbox(path: String, payload: [String: Any]) async throws {
+        _ = try await postInboxRaw(path: path, payload: payload)
+    }
+
+    /// Internal worker for `postInboxJSON` / `postInbox`. Builds and fires the request, maps
+    /// transport / non-2xx / non-HTTP failures to typed `ATTNError` cases, and returns the raw
+    /// response body for the caller to decode (or discard).
+    private func postInboxRaw(path: String, payload: [String: Any]) async throws -> Data {
         guard let url = URL(string: Self.inboxHost + path) else {
             Loggers.network.error("Invalid inbox URL for path \(path, privacy: .public)")
             throw ATTNError.badURL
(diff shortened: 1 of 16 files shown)


Sources/Inbox/InboxView.swift
```diff
@@ -27,6 +27,12 @@ struct InboxView: View {
                 buildListView {
                     ForEach(messages) { message in
                         InboxMessageRowView(message: message, style: viewModel.style)
+                            .contentShape(Rectangle())
+                            .onTapGesture {
+                                // Fires click tracking + local read flip. The SDK deliberately
+                                // does not open `actionURL` — host apps handle deep-link routing.
+                                viewModel.click(message.id)
```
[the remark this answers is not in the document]
adelag-attentive: Addressed in 25cdcd6. Added `Notification.Name.ATTNSDKInboxMessageTapped`; `InboxViewModel.click` now broadcasts it with `userInfo` carrying `attentiveInboxMessageId` (always) and `attentiveInboxActionUrl` (when the message has a valid actionURL). Mirrors the existing `ATTNSDKDeepLinkReceived` pattern used for push opens so hosts can register one observer for both entry points. 3 new InboxViewModelTests cover with-URL, without-URL, and malformed-URL cases.

Sources/API/ATTNAPI.swift
```diff
@@ -426,6 +426,27 @@ final class ATTNAPI: ATTNAPIProtocol {
...
+        actionURL: String?
+    ) async throws {
+        Loggers.network.debug("Reporting inbox click - Visitor ID: \(visitorId, privacy: .public), Message ID: \(messageId, privacy: .public)")
+        var payload: [String: Any] = [
+            "c": domain,
+            "visitor_id": visitorId,
+            "message_id": messageId
+        ]
+        if !pushToken.isEmpty { payload["push_token"] = "apns:\(pushToken)" }
+        if let actionURL = actionURL, !actionURL.isEmpty {
```
usharif-attentive, line 442: **Blocking: the server returns 400 when `action_url` is blank, but the SDK still fires the POST for messages without one.**

The backend handler (`mobile-api`'s `InboxMessageController.trackMessageClicked`) rejects the request outright:

```java
if (StringUtils.isBlank(request.actionUrl())) {
    return ResponseEntity.status(HttpStatus.BAD_REQUEST).build();
}
```

Here the key is omitted when nil/empty, and `InboxManager.markClicked` calls the API regardless of whether the message has an `actionURL` — so every tap on a message without an action URL is a guaranteed 400, silently swallowed by the fire-and-forget error handling. The click is never tracked and nothing surfaces it.

`testMarkClicked_omitsNilActionURL` / `testMarkClicked_omitsEmptyActionURL` codify this as correct because the mock doesn't validate.

Suggestion: skip the POST entirely when `actionURL` is blank (and test that), or get the backend contract relaxed first.

Sources/Inbox/InboxManager.swift
```diff
@@ -350,6 +350,37 @@ actor InboxManager {
...
+        let identity = identityProvider()
+        guard !identity.visitorId.isEmpty else {
+            Loggers.network.debug("Skipping inbox click tracking: empty visitor ID")
+            return
+        }
+
+        // Local read flip mirrors what markRead does; kept in-line so it always runs even if the
+        // network call throws.
+        if message.isRead == false {
+            messagesByID[messageID]?.isRead = true
```
usharif-attentive: **Blocking: the local read flip desyncs from the server and from the unread badge.**

This flips `isRead` locally but never tells the server — and the backend's `trackMessageClicked` is analytics-only; it does not mark the message read. So the unread dot clears on tap, then reverts on the next fetch or app launch.

The doc comment's "when the mark-read endpoint lands, that call will POST separately" is stale: `develop` already has `markRead` wired to `api.markMessagesRead`.

Also, `storedUnreadCount` — the badge value, reconciled only through `applyReadStatusResponse` — is never decremented here, so even locally the badge and row state disagree.

Suggestion: delegate the read side to the existing `markRead(messageID)` (which gets the optimistic flip, server POST, generation guarding, and count reconciliation for free) and keep the click POST as pure fire-and-forget.

Sources/API/ATTNAPI.swift
```diff
@@ -426,6 +426,27 @@ final class ATTNAPI: ATTNAPIProtocol {
...
+    /// Reports a message click to the server. Body carries `c` (domain) + `visitor_id` +
+    /// `push_token` + `message_id` + `action_url`; server returns 204 No Content on success.
+    func markMessageClicked(
+        pushToken: String,
+        visitorId: String,
+        messageId: String,
+        actionURL: String?
+    ) async throws {
+        Loggers.network.debug("Reporting inbox click - Visitor ID: \(visitorId, privacy: .public), Message ID: \(messageId, privacy: .public)")
+        var payload: [String: Any] = [
```
usharif-attentive: **Payload duplication.** This hand-builds the `c` / `visitor_id` / `apns:`-prefixed `push_token` block that `inboxIdentityPayload` already encapsulates. Calling the helper with `email: nil, phone: nil` would keep the prefix logic in one place — a future prefix change currently needs two edits.

usharif-attentive (commented):
Left four inline comments: two blocking issues (server 400 on blank `action_url`, local read flip desyncing from the server/badge) and two smaller ones (payload duplication, a timing-based test).

Tests/TestCases/InboxManagerTests.swift
```diff
@@ -382,4 +382,118 @@ final class InboxManagerTests: XCTestCase {
...
+            api: apiSpy,
+            identityProvider: identityProvider(visitorId: "")
+        )
+        // Seed a message manually via the fetch path so `messagesByID` is non-empty. Since the
+        // init refresh short-circuits on empty visitorId, drive one directly.
+        apiSpy.stubbedInboxMessagesResponses = [
+            InboxResponse(messages: [makeMessage(id: "1")], nextPageToken: nil)
+        ]
+        // Give init a moment to short-circuit (no state transition), then attempt click.
+        try? await Task.sleep(nanoseconds: 50_000_000)
```
usharif-attentive: **Flaky test pattern.** This relies on a 50ms `Task.sleep` to let init settle. Timing-based waits are how flakes get born — prefer awaiting an observable state, as the other tests do with `waitForLoadedState`.

Sources/API/ATTNAPI.swift
adelag-attentive: Addressed in 1b8a390. `markClicked` now skips the API POST entirely when `actionURLString` is nil/empty/whitespace, so the guaranteed 400 never leaves the device. Test `testMarkClicked_blankActionURL_skipsClickPostButStillMarksRead` covers the nil case; `testMarkClicked_emptyActionURLString_skipsClickPost` covers whitespace-only. The read side still runs so the row's dot + badge stay in sync.

Sources/Inbox/InboxManager.swift
adelag-attentive: Addressed in 1b8a390. Rebased onto latest develop (MSDK-410 landed) and dropped the open-coded local flip — `markClicked` now calls `markRead(id)`, so the click path inherits the server PATCH, generation guarding, and authoritative `unread_count` reconciliation via `applyReadStatusResponse`. Row dot + badge count now agree on the next fetch. Test `testMarkClicked_flipsLocalReadStateAndFiresMarkRead` asserts the delegation (checks `markMessagesReadCallCount == 1`).

Sources/API/ATTNAPI.swift
adelag-attentive: Addressed in 1b8a390. `markMessageClicked` now builds the payload via `inboxIdentityPayload(pushToken:, email: nil, phone: nil, visitorId:)`, so the `apns:` prefix + empty-token omission live in one place.

Tests/TestCases/InboxManagerTests.swift
adelag-attentive: Addressed in 1b8a390. Test now seeds via a real init fetch with a non-empty visitor (waiting on `waitForLoadedState` — same pattern as sibling tests), then flips identity to empty via a mutable-string reference (`MutableString` helper at the bottom of the file, `@unchecked Sendable`). No timing-based waits.

usharif-attentive (approved):

adelag-attentive:
backend code is still WIP per this slack thread, but i'm going to ship this to unblock e2e testing. https://attentivemobile.slack.com/archives/C09R7K58SSW/p1784661288520179

Will rebase and merge now (might bypass rules just for today if needed because i wanted to get both PRs in) @usharif-attentive

usharif-attentive (approved):

PR #278, opened 2026-07-21
Merge main into develop

usharif-attentive:
## Summary

Tracking PR to sync `develop` with `main` (39 commits, including releases 2.0.14 and 2.0.15).

### Notable changes coming from main
- **MSDK-312**: thread-safety hardening — lock-protected shared state in core SDK classes, `withLock` helper, concurrency test stabilization
- **MSDK-344**: `pushEnabled` init option; opt-in/out sent immediately for non-push clients; marketing-subscription code extracted to `ATTNSDK+MarketingSubscriptions.swift`
- **MSDK-354**: fix plus-sign encoding in URL query parameters
- **MSDK-315**: standardized error handling (`CustomNSError` conformance on `ATTNError`)
- **MSDK-314**: centralized hardcoded URLs, timeouts, and config values
- Bonni app: visitor-ID copy button (MSDK-370), `pushEnabled` toggle (MSDK-371), dark-mode support
- Docs/CI: CLAUDE.md, AGENTS.md, README updates, SwiftLint exclusions

### Conflict resolutions
- `ATTNSDK.swift`: kept main's `stateLock`-based `consumeDeepLink` and relocated marketing-subscription code; removed develop's inline duplicates; kept develop's `identityStore` and debug-log-capture API
- `ATTNError.swift`: union — develop's inbox error cases + main's `initializationFailed` and `CustomNSError` extension (inbox cases assigned codes 7/8)
- `SettingsViewController.swift`: kept develop's log-overlay switch alongside main's copy-visitor-ID button and push/v2 toggle handlers
- `ProductViewController.swift`: kept develop's inbox bar button
- `.swiftlint.yml` / `project.pbxproj`: additive union of both sides

## Verification
- `attentive-ios-sdk-framework` scheme builds for iOS Simulator
- Bonni `AttentiveExample` app builds for iOS Simulator
- No remaining conflict markers; unit tests deferred to CI


Sources/Public/SDK/ATTNSDK+MarketingSubscriptions.swift
```diff
@@ -0,0 +1,300 @@
...
+            Loggers.event.error("updateUser: aborted — missing push token - Tried in-memory token: '\(trimmedPushToken, privacy: .public)', Tried UserDefaults: '\(UserDefaults.standard.string(forKey: ATTNSDKConfiguration.UserDefaultsKey.deviceToken) ?? "nil", privacy: .public)', Visitor ID: \(self.userIdentity.visitorId, privacy: .public)")
+            callback?(nil, nil, nil, ATTNSDKError.missingPushToken)
+            return
+        }
+        Loggers.event.debug("updateUser: proceeding with push token: \(pushToken, privacy: .public)")
+        clearUserIdentifiers()
+        var newIdentifiers: [String: Any] = [:]
+        if let email = email { newIdentifiers[ATTNIdentifierType.email] = email }
+        if let phone = phone { newIdentifiers[ATTNIdentifierType.phone] = phone }
+        userIdentity.mergeIdentifiers(newIdentifiers)
```
adelag-attentive, line 170: **HIGH — `updateUser` never re-publishes the identity snapshot after merging new identifiers**

`updateUser(email:phone:)` calls `clearUserIdentifiers()` on line 158 (which internally publishes a *cleared* snapshot to `identityStore`) and then `userIdentity.mergeIdentifiers(newIdentifiers)` here to write the new email/phone — but never calls `publishIdentitySnapshot()` after the merge. `identify(_:)` and `clearUserIdentifiers()` both publish after their mutation; this path is the odd one out.

**Failure scenario:**
```swift
sdk.updateUser(email: "new@example.com", phone: "+15551234567")
// later
sdk.inbox.fetchUnreadCount()  // reads identity from identityStore
```
The inbox fetch runs with `email=nil, phone=nil` (the state left after `clearUserIdentifiers()`), not `new@example.com`. Server identity resolution then uses only `visitor_id + push_token`, so the count corresponds to the guest/logged-out inbox rather than the newly-identified user.

**Root cause:** merge artifact. Main’s `updateUser` called `mergeIdentifiers` but had no `identityStore`; develop’s `updateUser` had the identity store but did not call `mergeIdentifiers`. The merge combined both mutations without adding the `publishIdentitySnapshot()` that ties them together.

**Fix:** one line — call `publishIdentitySnapshot()` immediately after `userIdentity.mergeIdentifiers(newIdentifiers)`.

Sources/Public/SDK/ATTNSDK+MarketingSubscriptions.swift
```diff
@@ -0,0 +1,300 @@
...
+                callback: callback,
+                createdAt: Date()
+            ))
+            return
+        }
+
+        Loggers.event.debug("Processing opt-in marketing subscription - Visitor ID: \(self.userIdentity.visitorId, privacy: .public), Push Token: \(self.currentPushToken, privacy: .public), Email: \(email ?? "nil", privacy: .public), Phone: \(phone ?? "nil", privacy: .public)")
+
+        api.sendOptInMarketingSubscription(
+            pushToken: currentPushToken,
```
adelag-attentive: **HIGH — Stale APNs token leaks into opt-in/opt-out for `pushEnabled == false` after an upgrade**

The immediate-send path passes `currentPushToken` unchanged. `currentPushToken` reads from `PushTokenStore`, which is backed by `UserDefaults` at key `attentiveDeviceToken`. If an app previously ran with `pushEnabled: true` and stored a token there, then upgrades to `pushEnabled: false`, the observer that would clear the token (`handleAppDidBecomeActive`) is gated on `pushEnabled` in `ATTNSDK.init` and never registers. `currentPushToken` keeps returning the stale token, and it lands on `sendOptInMarketingSubscription` — attaching `pt=<stale apns>` + `tp=apns` to an install the SDK does not own for push. Same issue on line 107 for `sendOptOutMarketingSubscription`.

**Failure scenario:** app v1 shipped with `pushEnabled: true`, user granted push, token persisted to `UserDefaults`. App v2 flips to `pushEnabled: false`. First opt-in in v2 sends `POST /opt-in-subscriptions?pt=<v1 token>&tp=apns` — a token routed to an app that no longer expects pushes from Attentive.

**Fix:** clamp to empty string when push is disabled:
```swift
let pushTokenToSend = pushEnabled ? currentPushToken : ""
api.sendOptInMarketingSubscription(pushToken: pushTokenToSend, ...)
```
Alternately, wipe `UserDefaults[attentiveDeviceToken]` in `init` when `pushEnabled == false`. Note that `testOptIn_whenPushDisabled_sendsImmediatelyWithoutPushToken` passes only because `setUp()` clears that key manually — the real-world upgrade path is untested.

Sources/Public/SDK/ATTNSDK+MarketingSubscriptions.swift
usharif-attentive: Fixed in fb09df7 — `updateUser` now calls `publishIdentitySnapshot()` immediately after `userIdentity.mergeIdentifiers(newIdentifiers)`, so the identity store sees the new email/phone instead of the cleared state. (`publishIdentitySnapshot` moved from the fileprivate helpers extension to an internal one so it's callable from `ATTNSDK+MarketingSubscriptions.swift`.)

Sources/Public/SDK/ATTNSDK+MarketingSubscriptions.swift
usharif-attentive: Fixed in fb09df7 using the clamp approach: both opt-in and opt-out immediate-send paths now send `pushEnabled ? token : ""`. Went with clamping rather than wiping the UserDefaults key in `init` so hosts that toggle `pushEnabled` back on don't lose the persisted token. Also added regression tests that seed a stale `attentiveDeviceToken` before constructing a push-disabled instance and assert an empty token is sent. Note this bug originates on `main` (MSDK-344), so it exists there until the next release-back merge — flagging in case we want a ticket.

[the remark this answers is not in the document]
usharif-attentive: Fixed in fb09df7 — same finding as the review comment below from @adelag-attentive: both opt-in and opt-out paths now clamp the token to empty when `pushEnabled == false`, with regression tests covering the stale-persisted-token case.

Sources/Helpers/Extension/ATTNSDK+Extension.swift
```diff
@@ -9,9 +9,79 @@ import Foundation
...
+            return
+        }
+
+        if let addToCart = event as? ATTNAddToCartEvent {
+            guard !addToCart.items.isEmpty else {
+                Loggers.event.debug("No items found in the AddToCart event, skipping v2 send.")
+                return
+            }
+            for item in addToCart.items {
+                sendAddToCartEvent(product: product(from: item), currency: item.price.currency)
```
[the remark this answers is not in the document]
usharif-attentive: This was reviewed against the previous head (6686a01) and has since been fixed: the branch now includes MSDK-441 (#286) from `main`, which carries the legacy event's deeplink through `sendAddToCartEvent`/`sendProductViewEvent` into the v2 request (`eventRequest.deeplink = deeplink` in `ATTNSDK+Extension.swift`).

adelag-attentive (approved):

PR #279, opened 2026-07-24
MSDK-434: replace live incrementing inbox timestamp with static rounded relative time

usharif-attentive:
## Summary

Replaces the inbox message row's live-incrementing timestamp (SwiftUI's self-updating `.relative` text style) with a static rounded relative time, matching Android. Resolves [MSDK-434](https://attentivemobile.atlassian.net/browse/MSDK-434).

## Key Changes

- Timestamps now render as "Just now" / "31m ago" / "5h ago" / "3d ago", falling back to an absolute date ("Jan 27") for messages a week or older — the same bucketing Android uses. The value is computed once per render via `Calendar.dateComponents` rather than a timer, so it holds still while the inbox is visible and refreshes when the inbox is re-presented.
- Because the math is calendar-aware, values near a DST transition may differ from Android's raw-millisecond division by an hour (e.g. "1d ago" vs "23h ago") — deliberate trade-off for calendar-correct semantics.
- Strings go through `String(localized:)`, consistent with the rest of the Inbox module.
- Drive-by: removes a dead duplicate `.foregroundColor(.gray)` on the timestamp text, and commits the re-resolved workspace `Package.resolved`, which had been stale since the snapshot-testing dependency was added and gets rewritten by any workspace build.

## Public API Impact

None — the formatter is an internal `Date` extension; no public surface changes.

## Verification

- New `InboxDateTests` (Swift Testing, 13 parameterized cases) cover every bucket and its boundaries (59s → "Just now", 3599s → "59m ago", 23h59m → "23h ago", 6d23h → "6d ago", exactly 7d → absolute date, future timestamps → "Just now"), with a fixed injected clock and UTC calendar for determinism.
- Full unit suite passes: 268 tests, iPhone 17 simulator.
- Verified visually in the Example app: inbox timestamps render statically and no longer tick while the screen is open.

## Risk / Rollback

Low — internal, view-layer-only change to one label; no API, network, persistence, or threading changes. Purely visual regression surface. No kill-switch; rollback is a follow-up release, acceptable given the blast radius.

[MSDK-434]: https://attentivemobile.atlassian.net/browse/MSDK-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Inbox/Inbox+Date.swift
+++ b/Sources/Inbox/Inbox+Date.swift
@@ -1,0 +1,28 @@
+//
+//  Inbox+Date.swift
+//  attentive-ios-sdk
+//
+//  Created by Umair Sharif on 7/23/26.
+//
+
+import Foundation
+
+extension Date {
+    func inboxRelativeString(now: Date = Date(), calendar: Calendar = .current) -> String {
+        guard self <= now else { return String(localized: "Just now") }
+
+        let elapsed = calendar.dateComponents([.day, .hour, .minute], from: self, to: now)
+        switch (elapsed.day ?? 0, elapsed.hour ?? 0, elapsed.minute ?? 0) {
+        case (7..., _, _):
+            return formatted(.dateTime.month(.abbreviated).day())
+        case (let days, _, _) where days >= 1:
+            return String(localized: "\(days)d ago")
+        case (_, let hours, _) where hours >= 1:
+            return String(localized: "\(hours)h ago")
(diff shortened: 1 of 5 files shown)


usharif-attentive:
@codex review

usharif-attentive:
@attentive-claude review

adelag-attentive (approved):

PR #282, opened 2026-07-26
MSDK-439: keep inbox unread badge in sync via optimistic updates

adelag-attentive:
## Summary
- Optimistic ±1 unread-count adjustments on `markRead`/`markUnread` alongside the row-dot flip; rollback on PATCH failure. Delete already had this; brings the mutation methods to parity.
- `resetForIdentityChange` now spawns a background unread-count re-fetch for the new identity so the badge doesn't sit at 0 until the host next opens the inbox.
- Server value from `refreshUnreadCount()` and from the mark-read/unread PATCH response's `unread_count` field still fully replaces the cached count. The count is never derived from the loaded message list — pagination means the list is not authoritative for total unread.

## Ticket
[MSDK-439](https://attentivemobile.atlassian.net/browse/MSDK-439)

## Test plan
- [x] `InboxManagerTests` — 49 tests pass, including 9 new ones covering optimistic decrement/increment mid-flight, failure revert, zero-clamp, no-op on already-matching state, reset re-fetch behavior, and empty-visitor-id skip.
- [ ] Manual: reproduce the ticket's scenario in the Bonni demo app — tap unread → badge decrements immediately (not after PATCH), swipe mark-unread → badge increments immediately, delete unread → badge decrements, `clearUser()`/`updateUser()` → badge shows the new user's count without opening the inbox.
- [ ] Manual: verify the video's initial mismatch (badge 5 vs 4 visible dots at t=0). Note: that specific case is out of scope for this SDK fix — it originates in the server unread-count endpoint diverging from the message list's `isRead` flags. If the mismatch reproduces on a fresh open with no local mutation, split a backend ticket.

[MSDK-439]: https://attentivemobile.atlassian.net/browse/MSDK-439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Inbox/InboxManager.swift
+++ b/Sources/Inbox/InboxManager.swift
@@ -73,6 +73,15 @@ actor InboxManager {
     /// superseded by a newer in-flight call or an identity reset.
     private var refreshGeneration: UInt = 0
 
+    /// Monotonic counter bumped every time `storedUnreadCount` is written from an authoritative
+    /// source (server count fetch, mark-read/unread PATCH response, or identity reset). Snapshot
+    /// at the start of a mark-read/unread and re-checked on its rollback path so an overlapping
+    /// mutation's authoritative count is not clobbered by a stale ±1 revert. Distinct from
+    /// `refreshGeneration`, which only advances on count fetches and identity resets — a peer
+    /// mark-read/unread response reconciles the count without bumping that generation and would
+    /// otherwise slip past the existing guard.
+    private var unreadCountRevision: UInt = 0
+
     /// Monotonic counter used to discard responses from message refresh/pagination calls that
     /// have been superseded by a newer identity or a newer top-level `refresh()`.
     private var messagesGeneration: UInt = 0
@@ -330,6 +339,7 @@ actor InboxManager {
             guard generation == refreshGeneration else { return }
             let previous = storedUnreadCount
             storedUnreadCount = count
+            unreadCountRevision &+= 1
             // Re-emit only when the value actually changed and we're in a loaded state — avoids
             // waking subscribers with an identical payload on every no-op refresh.
             if !skipNotify, count != previous, case .loaded = currentState {
@@ -364,8 +374,11 @@ actor InboxManager {
         // Snapshot the refresh generation before the request. If an identity change
         // (`resetForIdentityChange`) or a newer `refreshUnreadCount` lands while the PATCH is in
         // flight, the generation advances and we drop this now-stale response instead of
-        // restoring an old unread count.
+        // restoring an old unread count. `unreadCountRevision` also catches a peer mark-read/
+        // unread whose response reconciled the count without bumping `refreshGeneration` — the
+        // count is now authoritative on that peer's return and this rollback must not clobber it.
         let generation = refreshGeneration
+        let revisionAtStart = unreadCountRevision
 
         do {
             let response = try await api.markMessagesRead(
@@ -380,7 +393,10 @@ actor InboxManager {
             guard generation == refreshGeneration else { return }
             if messagesByID[messageID] != nil, !previousIsRead {
                 messagesByID[messageID]?.isRead = previousIsRead
-                if didDecrementUnreadCount {
+                // Per-message state (`isRead`) is always reverted — the row must reflect this
+                // message's true state. Count is only reverted if no other authoritative write
+                // landed in between; otherwise the peer already wrote the correct total.
+                if didDecrementUnreadCount, revisionAtStart == unreadCountRevision {
                     storedUnreadCount += 1
                 }
                 send(.loaded(orderedMessagesSnapshot()))
@@ -407,11 +423,10 @@ actor InboxManager {
             send(.loaded(orderedMessagesSnapshot()))
         }
 
-        // Snapshot the refresh generation before the request. If an identity change
-        // (`resetForIdentityChange`) or a newer `refreshUnreadCount` lands while the PATCH is in
-        // flight, the generation advances and we drop this now-stale response instead of
-        // restoring an old unread count.
+        // See `markRead` for why the count-revision snapshot is separate from the refresh
+        // generation.
         let generation = refreshGeneration
+        let revisionAtStart = unreadCountRevision
 
         do {
             let response = try await api.markMessagesUnread(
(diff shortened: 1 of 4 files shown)


Sources/Inbox/InboxManager.swift
```diff
@@ -514,6 +549,9 @@ actor InboxManager {
         if case .loaded = currentState {
             send(.loaded([]))
         }
+        Task { [weak self] in
+            await self?.performUnreadCountFetch(skipNotify: false)
+        }
```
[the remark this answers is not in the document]
adelag-attentive: Addressed in 04cbb87. Moved the re-fetch out of `resetForIdentityChange` and onto ATTNSDK's `/user-update` callback (both `clearUser` and `updateUser`). Reset is now purely local, so the fetch runs only after the server-side visitor→email/phone association completes — no more caching the anonymous visitor's count for a user that's about to be re-identified. New test: `testResetForIdentityChange_followedByRefreshUnreadCount_populatesNewCount`.

Sources/Inbox/InboxManager.swift
```diff
@@ -361,6 +380,9 @@ actor InboxManager {
             guard generation == refreshGeneration else { return }
             if messagesByID[messageID] != nil, !previousIsRead {
                 messagesByID[messageID]?.isRead = previousIsRead
+                if didDecrementUnreadCount {
+                    storedUnreadCount += 1
```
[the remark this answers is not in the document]
adelag-attentive: Addressed in 04cbb87. Added `unreadCountRevision` that bumps on every authoritative count write (server count fetch, mark-read/unread PATCH response, identity reset). `markRead`/`markUnread` snapshot it alongside `refreshGeneration`; the failure-rollback still restores `isRead` unconditionally (that's per-message truth) but only reverts the count when the revision hasn't moved. New test: `testMarkRead_failure_afterPeerAuthoritativeWrite_doesNotClobberCount` — two overlapping markRead Tasks (as `InboxViewModel` schedules them), peer succeeds first, primary fails; asserts the peer's authoritative unread_count survives the primary's rollback.

Sources/Inbox/InboxManager.swift
```diff
@@ -497,12 +544,18 @@ actor InboxManager {
...
+    /// visitor→email/phone association *after* this reset, so a fetch fired now would run
+    /// against an unlinked anonymous visitor and cache the wrong count. Callers (see
+    /// `ATTNSDK.clearUserIdentifiers`) should invoke `refreshUnreadCount()` from the
+    /// `/user-update` callback instead — that guarantees the fetch sees the new identity.
     func resetForIdentityChange() {
         refreshGeneration &+= 1
         messagesGeneration &+= 1
         messagesReplacedCount &+= 1
         storedUnreadCount = 0
+        unreadCountRevision &+= 1
```
usharif-attentive, line 558: **Race: post-reset `refreshUnreadCount()` can coalesce with a stale init fetch and skip the network entirely.**

If an identity change lands while the manager's init-time fetch is still undrained, the re-fetch fired from the `/user-update` callback hits the coalesce branch at the top of `refreshUnreadCount()`, drains the stale init task, and returns without issuing a request. The init fetch's result is (correctly) discarded by the generation guard bumped here — so the new identity's badge sits at 0 until the next explicit refresh. That's the exact symptom MSDK-439 is fixing, so shipping with this window open undermines the feature.

Suggested fix: clear `initialRefreshTask` here in `resetForIdentityChange()` — post-reset, 0 is the correct value for anyone awaiting it, and the follow-up `refreshUnreadCount()` will then actually fetch. (Alternatively, skip the coalesce when `refreshGeneration` has advanced since the init task was created.)

usharif-attentive (commented):
Two inline findings from review — one correctness race that can reproduce the MSDK-439 symptom, one thread-safety issue in the new callback wiring.

Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -959,6 +973,15 @@ fileprivate extension ATTNSDK {
         return manager
     }
 
+    /// Fires a background inbox unread-count refresh only when the manager already exists.
+    /// Called from the `/user-update` callback path after an identity change so the badge picks
+    /// up the newly-associated user's server count. Skips materialization so host apps that
+    /// never touch the inbox don't pay for a network call on every `clearUser`/`updateUser`.
+    func refreshInboxUnreadCountForNewIdentityIfMaterialized() {
+        guard let manager = _inboxManager else { return }
```
usharif-attentive: **Data race: `_inboxManager` is read off the main thread here.**

`ATTNAPI.updateUser` invokes its callback directly on the URLSession completion queue, so this helper runs on a background thread — reading the plain `private var _inboxManager` while `materializedInboxManager()` writes it from the main thread. TSan and Swift 6 strict concurrency will both flag this.

Side effect of the same wiring: the host's chained `updateUser` callback now also runs on that background queue, which is a subtle behavior change for integrators that touch UI in the callback.

Suggest hopping to the main actor (or guarding `_inboxManager` with the same lock pattern as `identityStore`) before touching SDK state.

Sources/Inbox/InboxManager.swift
adelag-attentive: Good catch, addressed in a2e468d. `resetForIdentityChange` now sets `initialRefreshTask = nil` so the follow-up `refreshUnreadCount()` from the `/user-update` callback skips the coalesce branch and actually posts. The stale init task's response is still discarded by the generation guard when it eventually returns, and post-reset `storedUnreadCount = 0` so any `awaitInitialRefresh()` reader is served correctly by the nil short-circuit. New test: `testResetForIdentityChange_duringInitFetch_followupRefreshStillPostsForNewIdentity` — uses a new `onFetchInboxUnreadCount` spy hook to freeze the init fetch inside its stub, resets mid-flight, then verifies the follow-up refresh issues its own POST and lands the new identity's count.

Sources/Public/SDK/ATTNSDK.swift
adelag-attentive: Addressed the data race in a2e468d — the helper now hops to `DispatchQueue.main.async` before reading `_inboxManager`, matching the write-side thread contract.

On the host callback behavior side: `ATTNAPI.updateUser`'s callback was already invoked on the URLSession completion queue prior to this PR (see `ATTNAPI.swift:361`, inside the `urlSession.dataTask` completion block). The chained callback added by this PR just calls the host's callback on that same queue and then triggers the refresh — so integrators aren't seeing a threading-contract change, only a new call sink. Let me know if you'd prefer we also main-hop the host callback while we're here.

usharif-attentive (approved):

PR #283, opened 2026-07-27
MSDK-442: relax v2 event API for catalog-hydration clients

adelag-attentive:
## Summary

Two client-facing fixes plus one internal alignment on the `useV2Endpoint` / `recordEvent(_ eventData:)` path so Target's blockers on migrating off `/e` go away. Also retroactively fixes `cartTotal` on Carter's v2 Purchase events (MSDK-262 follow-up).

- **`ATTNProduct.name` → optional** (`String?`) with `encodeIfPresent`. Hosts using server-side catalog hydration can now send only `productId + price + quantity`. The backend schema doesn't require `name` and `ProductConverter.convertToDomainProductModel` null-checks it; the SDK type was the sole constraint.
- **`ATTNCart` gains `cartTotal` and `cartDiscount`.** On the auto-convert path the SDK now preserves `cart.cartTotal` / `cart.cartDiscount` when the host sets them, and falls back to an SDK-computed `cartTotal` (sum of item prices, using the legacy `priceFormatter`) when the host does not — matching what the legacy `/e` path emits. Previously the auto-convert dropped `cartTotal` to `null` on the wire.
- **Purchase `orderTotal` on the auto-convert path now matches the legacy `/e` formula** (`sum(price)`, quantity-agnostic) instead of `sum(price × quantity)`. Restores numeric continuity for hosts flipping `useV2Endpoint = true` and aligns iOS with Android (which never multiplied by quantity). A follow-up can revisit the correct cross-platform semantics as a coordinated change.

## Deferred (documented in the ticket, not in this PR)

Making `orderTotal` optional on `ATTNEventData.purchase(...)` with an SDK-computed default. Skipped as convenience-only scope creep; Target's flow explicitly passes it. Can be a follow-up when we hear it's actually needed.

## Public API impact

Additive and backward-compatible:

- `ATTNProduct(productId:name:price:quantity:)` — `name` argument becomes optional (nil default). Existing call sites passing a non-nil string keep compiling and behave the same.
- `ATTNCart` — two new `@objc public var` properties (`cartTotal`, `cartDiscount`). No existing initializer signature changes; hosts that never set them see the same on-wire behavior as before (with the added bonus that the auto-convert path now emits an SDK-computed `cartTotal` for them, populating a field that used to be `null`).

No public method signatures removed. No deprecations needed.

## Host-app rollback

Any host currently pinned to `2.0.15` can stay there — no wire-level breaking changes. Hosts adopting this version will see `cartTotal` populated on Purchase events when it previously arrived null. If a host was somehow depending on the (buggy) `null cartTotal` behavior, they can override by passing an explicit empty string on `ATTNCart.cartTotal`. Unlikely.

## Verification

- `xcodebuild test -scheme "attentive-ios-sdk Tests" -destination "platform=iOS Simulator,name=iPhone 16 Pro"` — 209 passed, 0 failed (iOS Simulator 26.1).
- SwiftLint: no new warnings introduced on changed files.
- New test coverage in `Tests/TestCases/ATTNV2EventModelsTests.swift` and `Tests/TestCases/ATTNSDKTests.swift`:
  - `ATTNProduct` init with nil `name` succeeds
  - JSON encoding omits `name` when nil
  - JSON encoding includes `name` when present
  - `useV2Endpoint = true` Purchase populates `cart.cartTotal` from items (legacy formula) when host provides a `cartId`-only `ATTNCart`
  - Host-provided `cart.cartTotal` / `cart.cartDiscount` win over the SDK-computed fallback
  - Missing cart entirely still produces a cart payload carrying the computed `cartTotal`
  - Existing multi-item `orderTotal` test asserts the new legacy value (`"15.50"` from `10.00 + 5.50`, not the old `"36.5"` from `10×2 + 5.5×3`)

## Test plan

- [ ] Bonni build with v2 endpoint toggle on, fire a Purchase with 2 items × quantity > 1, verify `cartTotal` and `orderTotal` in the `/mobile` request body match the pre-2.0.15 legacy values from `/e`
- [ ] Bonni Purchase with host-provided `ATTNCart(cartId:).cartTotal = "999.99"` — request body contains `"cartTotal": "999.99"` (not the SDK-computed value)
- [ ] Any host still on `useV2Endpoint = false` sees identical `/e` behavior (regression check)
- [ ] `recordEvent(.addToCart(product: ATTNProduct(productId:price:quantity:), currency:))` with no `name` produces a `/mobile` payload without a `"name"` key on the product

## Related

- MSDK-440 — Target endpoint / v2 questions
- MSDK-262 — Carter's `request_url` regression on `useV2Endpoint = true` (separate issue, shared code path; this PR also improves their Purchase `cartTotal` retroactively once they update)
- SUP-12358 — Target's `cartTotal` observation
- MSDK-337 — introduced the `useV2Endpoint` toggle and the auto-conversion path modified here

### DIFF
--- a/Sources/API/ATTNEventTypes.swift
+++ b/Sources/API/ATTNEventTypes.swift
@@ -53,10 +53,13 @@ public struct ATTNIdentifiers: Codable {
 }
 
 /// Represents an individual product in an event payload.
+///
+/// `name` is optional so hosts using server-side catalog hydration can send
+/// only `productId` and let the backend enrich the remaining fields.
 public struct ATTNProduct: Codable {
     public let productId: String
     public let variantId: String?
-    public let name: String
+    public let name: String?
     public let variantName: String?
     public let imageUrl: String?
     public let categories: [String]?
@@ -67,7 +70,7 @@ public struct ATTNProduct: Codable {
     public init(
         productId: String,
         variantId: String? = nil,
-        name: String,
+        name: String? = nil,
         variantName: String? = nil,
         imageUrl: String? = nil,
         categories: [String]? = nil,
@@ -90,7 +93,7 @@ public struct ATTNProduct: Codable {
         var container = encoder.container(keyedBy: CodingKeys.self)
         try container.encode(productId, forKey: .productId)
         try container.encode(variantId, forKey: .variantId)
-        try container.encode(name, forKey: .name)
+        try container.encodeIfPresent(name, forKey: .name)
         try container.encode(variantName, forKey: .variantName)
         try container.encode(imageUrl, forKey: .imageUrl)
         try container.encode(categories, forKey: .categories)
--- a/Sources/API/ATTNEventTypes.swift
+++ b/Sources/API/ATTNEventTypes.swift
@@ -54,12 +54,15 @@ public struct ATTNIdentifiers: Codable {
 
 /// Represents an individual product in an event payload.
 ///
-/// `name` is optional so hosts using server-side catalog hydration can send
-/// only `productId` and let the backend enrich the remaining fields.
+/// `name` has a default of `""` so hosts using server-side catalog hydration
+/// can omit it and let the backend enrich the field from the product catalog.
+/// The property type stays `String` (non-optional) to preserve source and ABI
+/// compatibility with hosts that read `product.name` as a non-optional value;
+/// an empty string is omitted from the wire payload.
 public struct ATTNProduct: Codable {
     public let productId: String
     public let variantId: String?
-    public let name: String?
+    public let name: String
     public let variantName: String?
     public let imageUrl: String?
     public let categories: [String]?
@@ -70,7 +73,7 @@ public struct ATTNProduct: Codable {
     public init(
         productId: String,
         variantId: String? = nil,
-        name: String? = nil,
+        name: String = "",
         variantName: String? = nil,
         imageUrl: String? = nil,
         categories: [String]? = nil,
@@ -93,7 +96,12 @@ public struct ATTNProduct: Codable {
         var container = encoder.container(keyedBy: CodingKeys.self)
         try container.encode(productId, forKey: .productId)
         try container.encode(variantId, forKey: .variantId)
-        try container.encodeIfPresent(name, forKey: .name)
+        // Omit the key entirely when the host didn't supply a name so the
+        // backend product-catalog hydration takes over. An empty string on the
+        // wire would suppress hydration and leave downstream systems with "".
+        if !name.isEmpty {
+            try container.encode(name, forKey: .name)
+        }
         try container.encode(variantName, forKey: .variantName)
         try container.encode(imageUrl, forKey: .imageUrl)
         try container.encode(categories, forKey: .categories)
--- a/Sources/Helpers/Extension/ATTNSDK+Extension.swift
+++ b/Sources/Helpers/Extension/ATTNSDK+Extension.swift
@@ -23,16 +23,23 @@ extension ATTNSDK {
                 return
             }
             let products = purchase.items.map { product(from: $0) }
-            let cart = ATTNCartPayload(from: purchase.cart)
             let currency = purchase.items[0].price.currency
-            let orderTotal = purchase.items.reduce(NSDecimalNumber.zero) { total, item in
-                let quantity = NSDecimalNumber(value: item.quantity)
-                return total.adding(item.price.price.multiplying(by: quantity))
+            // Match the legacy /e computation exactly (sum of item prices,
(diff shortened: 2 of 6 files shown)


Sources/API/ATTNEventTypes.swift
```diff
@@ -53,10 +53,13 @@ public struct ATTNIdentifiers: Codable {
...
 
 /// Represents an individual product in an event payload.
+///
+/// `name` is optional so hosts using server-side catalog hydration can send
+/// only `productId` and let the backend enrich the remaining fields.
 public struct ATTNProduct: Codable {
     public let productId: String
     public let variantId: String?
-    public let name: String
+    public let name: String?
```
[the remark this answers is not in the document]
adelag-attentive: I checked Apple's library-evolution docs after pushing back and "changing the type of a property" is explicitly prohibited under library evolution, and `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` is set on our XCFramework (Fastfile:286). Even setting XCFramework ABI aside, silently rolling every client onto v2 means we shouldn't break the source-level contract where a host writes `let s: String = product.name`. That was the real risk you flagged.

Reworked in 219fd33 without changing the property type:

- `ATTNProduct.name` stays `String` (not `String?`). Getter mangling is unchanged; existing hosts reading `product.name` as a non-optional keep compiling and keep binding to the same accessor.
- `init` now defaults `name: String = ""`. Existing keyword call sites (`ATTNProduct(productId:name:price:quantity:)`) still resolve to the same overload; new catalog-only callers can omit the arg. Default values are compiled at the call site so the callee's mangled symbol is stable.
- `encode(to:)` skips the `name` key when it's empty, so the wire-level goal (let backend product-catalog hydration take over) is met without exposing an empty string downstream.
- `product(from: item)` in the auto-convert path goes back to `item.name ?? ""` — matches the original pre-2.0.15 shape.

Net effect: zero source or ABI changes visible to existing consumers, catalog-hydration hosts can now omit `name`, and Target's blocker is removed. Tests updated (`testATTNProduct_init_nameOmitted_defaultsToEmpty`, `testATTNProduct_encode_omitsNameWhenEmpty`). Full test suite still green.

Sources/Helpers/Extension/ATTNSDK+Extension.swift
```diff
@@ -23,16 +23,23 @@ extension ATTNSDK {
...
-            let orderTotal = purchase.items.reduce(NSDecimalNumber.zero) { total, item in
-                let quantity = NSDecimalNumber(value: item.quantity)
-                return total.adding(item.price.price.multiplying(by: quantity))
+            // Match the legacy /e computation exactly (sum of item prices,
+            // quantity-agnostic, formatted with 2 fraction digits) so flipping
+            // useV2Endpoint doesn't silently change historical totals.
+            let computedTotalNumber = purchase.items.reduce(NSDecimalNumber.zero) { total, item in
+                total.adding(item.price.price)
             }
+            let computedTotal = purchase.priceFormatter.string(from: computedTotalNumber) ?? computedTotalNumber.stringValue
```
usharif-attentive, line 33: `priceFormatter` is a plain `NumberFormatter` with no locale pinning, so it formats with the device locale's decimal separator — on a German or French device this sends `orderTotal`/`cartTotal` as `"15,50"`, not `"15.50"`. The old v2 code used `NSDecimalNumber.stringValue`, which is locale-independent, so this change imports the legacy `/e` locale bug into the newer endpoint. If the `/mobile` backend parses these as decimals, non-US-locale devices will send unparseable totals.

Suggest either pinning the formatter to `en_US_POSIX` on both paths (arguably what "legacy value" was always intended to mean), or explicitly confirming the backend tolerates comma separators and noting it in the PR. Tests won't catch this — simulators run en_US.

adelag-attentive: Good catch — verified this fails against the backend. `ProductConverter.buildDomainCart` calls `Double.valueOf(cartTotal)` and `PurchaseProcessor.buildBillingTotals` uses `new BigDecimal(orderTotal)`; both throw `NumberFormatException` on `"15,50"`, which then falls through to the "Unable to parse order total" log path and the value is dropped.

Fixed in ad1c8b0 by pinning the shared `priceFormatter` to `en_US_POSIX`. Applies to both the legacy `/e` path (latent bug of the same shape) and the new v2 auto-convert. Also added `testPriceFormatter_pinsToPOSIX_regardlessOfDeviceLocale` — asserts `locale.identifier == "en_US_POSIX"` and that a rendered decimal never contains `,`, so a future unpinning regresses even on the en_US simulator.

usharif-attentive (approved):

PR #284, opened 2026-07-27
MSDK-439 (part 2): reconcile unread count against count endpoint after every mutation

adelag-attentive:
## Summary

Follow-up to #282 for [MSDK-439](https://attentivemobile.atlassian.net/browse/MSDK-439). Fixes the persistent symptom Umair reported: **after mark-read / mark-unread / delete in the inbox view, popping back to the host page leaves the inbox badge missing entirely (as if the count is 0). A second inbox visit + pop restores the correct count.**

## Root cause — confirmed backend bug ([CI-566](https://attentivemobile.atlassian.net/browse/CI-566))

`mobile-api`'s handlers for `PATCH /inbox/messages/read` and `PATCH /inbox/messages/unread` **hardcode `unread_count = 0` in their response DTO**, regardless of the user's actual unread state. Traced to:

- `mobile-api/src/main/java/com/attentivemobile/mobile/api/client/InboxMessageClientImpl.java` lines 70-71 and 99-100:
  ```java
  // Note: The actual unread count would need to be fetched separately or returned by the service
  return new UpdateReadStatusResult(statuses, 0);
  ```
- Upstream proto (`contracts/mobile-inbox-service/.../MobileInbox.proto`) also lacks the field on `MarkReadResponse` / `MarkUnreadResponse`, so the mobile-api layer had no source to read from — the `0` is a stub the endpoint owner never wired up.

This has been filed as **[CI-566](https://attentivemobile.atlassian.net/browse/CI-566)** with reporter Adela Gao. Preferred backend fix: add `unread_count` to the proto responses and wire it through so the SDK gets the real count from the PATCH round-trip.

## Why the SDK-side workaround is worth landing anyway

The bug affects every user of the iOS SDK v0.6.x+ **right now**. Even after CI-566 ships, the mobile-api rollout will take time, and older SDK versions in the field will still hit this until they update.

The defensive fix in this PR makes the SDK **immune to the field** — regardless of what mobile-api's PATCH returns in `unread_count` (0, wrong value, missing), the SDK never trusts it. The badge count is always sourced from `POST /inbox/messages/unread/count` (the endpoint the RFC declared source-of-truth).

Once CI-566 is fully deployed AND the SDK version depending on this workaround has aged out of the field, the defensive path could be simplified (trust the PATCH's `unread_count` again, drop the trailing count fetch). That's a follow-up decision, not required now.

## What this PR changes

`Sources/Inbox/InboxManager.swift`:

1. **`applyReadStatusResponse` no longer writes `storedUnreadCount = response.unreadCount`.** Only the per-message `isRead` echo is applied (direct request↔response mapping — trustworthy).
2. **Every `markRead` / `markUnread` / `delete` fires a background `/inbox/messages/unread/count` fetch** via `defer { scheduleUnreadCountReconciliation() }`. Runs on both success and failure paths so the badge converges to server truth in a bounded window regardless of what the PATCH returned.
3. **Optimistic ±1 stays** for immediate visual feedback between tap and reconciliation.
4. **Rollback logic split**: `isRead` reverts unconditionally on PATCH failure (per-message truth), count revert stays gated on `unreadCountRevision` (protects against peer-mutation clobber).

Doc comments call out CI-566 by ticket key and describe the tradeoff for a future maintainer.

## Why this fix is robust across host architectures

The bug reproduces regardless of how the host observes state — UIKit `viewWillAppear` cancelling stream subscriptions, SwiftUI `.task` on view lifecycle, Combine subscribers, or direct polling on `sdk.unreadCount`. The SDK now self-heals: the trailing count fetch lands within one network round-trip of any mutation, writing the correct value into `storedUnreadCount` before the host next reads it. No host-side "refresh on appear" hook is required.

References consulted (Apple):
- [SE-0314 (`AsyncStream`)](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0314-async-stream.md) — confirmed `AsyncStream` is single-consumer by design; buffering is unbounded by default; the build closure runs synchronously in the initializer on the caller's executor.
- [SE-0304 (Structured Concurrency)](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0304-structured-concurrency.md) — unstructured `Task {}` is the correct pattern for fire-and-forget reconciliation.

## Test plan

- [x] 59 InboxManager unit tests pass (up from 51 pre-PR), including 8 new ones for the trailing reconciliation on success/failure/no-op paths and a `testMarkRead_recoversFromWrongPatchUnreadCount` regression test that stubs `unread_count: 0` from the PATCH and asserts the badge converges to the count endpoint's value.
- [ ] Manual (once merged into a testable build): reproduce the ticket's scenario in the Bonni demo app — tap read/unread/delete, back → badge shows the correct count without needing a second visit.
- [ ] Verify no regression on the second-visit path — refresh + navigation back should still show correct count.

## Related tickets

- Parent (mobile SDK): [MSDK-439](https://attentivemobile.atlassian.net/browse/MSDK-439)
- Backend fix ticket: [CI-566](https://attentivemobile.atlassian.net/browse/CI-566)
- Preceding SDK PR: #282 (commit `3067351` on `develop`)

[MSDK-439]: https://attentivemobile.atlassian.net/browse/MSDK-439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CI-566]: https://attentivemobile.atlassian.net/browse/CI-566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CI-566]: https://attentivemobile.atlassian.net/browse/CI-566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Inbox/InboxManager.swift
+++ b/Sources/Inbox/InboxManager.swift
@@ -360,10 +360,10 @@ actor InboxManager {
         guard let identity = requireIdentity(context: "inbox mark-read") else { return }
 
         // Optimistically decrement the badge alongside the row-dot flip so the two agree
-        // between the tap and the PATCH response. The PATCH's `unread_count` is a hint that
-        // narrows the reconciliation window; the follow-up count-endpoint fetch (spawned in
-        // `defer`) is what actually lands the authoritative value. Clamping at 0 keeps a
-        // stale-local case from producing a negative badge.
+        // between the tap and the trailing count fetch. The PATCH's `unread_count` response
+        // field is ignored (see `applyReadStatusResponse` and CI-566); the follow-up count-
+        // endpoint fetch (spawned in `defer`) is what actually lands the authoritative value.
+        // Clamping at 0 keeps a stale-local case from producing a negative badge.
         var didDecrementUnreadCount = false
         if !previousIsRead {
             messagesByID[messageID]?.isRead = true
@@ -374,13 +374,9 @@ actor InboxManager {
             send(.loaded(orderedMessagesSnapshot()))
         }
 
-        // Reconcile the count against the authoritative endpoint after every mutation,
-        // success or failure. The PATCH's `unread_count` and the optimistic ±1 can both
-        // drift — from mobile-api DTO quirks, from mid-flight peer mutations, or from a
-        // clamped optimistic path — and the host may not have an observer active when
-        // this method returns (e.g., the user navigates back before the PATCH settles).
-        // A trailing count fetch guarantees the badge converges to server truth without
-        // requiring any host-side "refresh on appear" hook.
+        // Reconcile against the count endpoint after every mutation. See
+        // `scheduleUnreadCountReconciliation` for the full rationale (backend gap CI-566 + host
+        // lifecycle robustness).
         defer { scheduleUnreadCountReconciliation() }
 
         // Snapshot the refresh generation before the request. If an identity change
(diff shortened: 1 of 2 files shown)


adelag-attentive:
Closing pre-release: this workaround is unnecessary because MSDK-439's actual root cause is backend bug [CI-566](https://attentivemobile.atlassian.net/browse/CI-566) (mobile-api's mark-read/mark-unread endpoints hardcode `unread_count: 0`). Since we're still pre-release, no shipped SDK version is affected, so there's no value in shipping SDK-side compensation.

If CI-566 slips past the iOS SDK GA date, or an early-access integrator hits the bug before it ships, reopen this PR. Otherwise the story closes with zero SDK code changes once CI-566 lands and is verified in the Bonni test app.

[CI-566]: https://attentivemobile.atlassian.net/browse/CI-566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #285, opened 2026-07-28
MSDK-213: open push and inbox deeplinks on tap

usharif-attentive:
## Summary

Implements MSDK-213: the SDK now opens deeplinks on the host app's behalf when the user taps a push notification or an inbox message, matching the Android SDK's behavior. Previously both surfaces only broadcast the URL and left navigation entirely to the host.

## Key Changes

**Push notifications** — a tapped push carrying `attentive_open_action_url` now opens the link after the existing `ATTNSDKDeepLinkReceived` broadcast. Custom-scheme URLs route into the host app's URL handlers; `http(s)` URLs are opened with `.universalLinksOnly` so the user is never bounced to the browser. Hosts that already navigate on their own can set `automaticallyOpensPushDeepLinks = false`.

**Inbox messages** — tapping a row in the built-in inbox UI fires click tracking first, posts the existing `ATTNSDKInboxMessageTapped` broadcast, then opens the message's `actionURL`. Unlike push, unclaimed `http(s)` links here fall back to the browser (per the ticket's requirement). Two opt-outs: `automaticallyOpensInboxDeepLinks = false` suppresses only the SDK-initiated open (tracking + broadcast still fire), or pass an `onMessageTap` closure to `inboxView()` / `inboxViewController()` to replace the default routing entirely — it receives every tap (even URL-less messages, matching Android's `onMessageClick` override).

**URL validation** (revised per review): both paths broadcast **every URL that parses with a non-empty scheme** — host visibility is never gated on validation. Only the SDK-initiated open is gated: scriptable schemes (`javascript:`, `file:`, `data:`, `about:`, `vbscript:`) and privileged system-action schemes (`tel:`, `telprompt:`, `sms:`, `mailto:`, `facetime:`, `facetime-audio:`, `itms`/`itms-apps`/`itms-services:`) are never handed to `UIApplication`. The scheme-shape check deliberately admits non-RFC shapes (`my_app://`, `1password://`) that the lenient pre-iOS 17 parser accepts on iOS 15/16. URL opening goes through a new internal `ATTNURLOpening` protocol so both paths are unit-testable.

**Docs** — README gains an upgrade warning for hosts that already navigate off the broadcast/`consumeDeepLink()`; CHANGELOG gains an Unreleased section covering the behavior change and both opt-outs.

## Public API Impact

Additive and source-compatible: two new `@objc` public properties (`automaticallyOpensPushDeepLinks`, `automaticallyOpensInboxDeepLinks`, both default `true`) and two new optional parameters (`onMessageTap` on the inbox entry points). **Both defaults change existing runtime behavior**: hosts that navigate via the broadcasts today will see the SDK also open the URL unless they opt out. Inbox is pre-GA (alpha: JTV/Aero/JCP), so the blast radius there is limited; the push behavior change applies to push integrations adopted since the Oct 2025 push beta (when the broadcast shipped). Android precedent: SDK-driven navigation is Android's only push behavior (no broadcast alternative exists), and Android's inbox defaults to SDK-opens with an `onMessageClick` override — this PR converges iOS on that contract. Default-on vs. opt-in is under discussion in review threads.

## Verification

Unit tests only so far: full suite green on iPhone 17 simulator (iOS 26.5), including 11 `InboxViewModelTests` (default open, browser-fallback options, open-failure logging, custom-handler bypass, nil-URL, broadcast shape, auto-open opt-out, flag read at tap time) and 12 push/URL-validation tests in `ATTNSDKTests` (custom scheme, universal-links-only, whitespace, opt-out, invalid URL, empty scheme, blocked-scheme broadcast-but-never-open for scriptable and privileged schemes, non-RFC scheme acceptance, broadcast, consume). The plan's manual Bonni checks are still pending: universal link resolving into the app, Safari fallback, unclaimed custom scheme, and a custom `onMessageTap` integration.

## Risk

Medium. The main risk is double navigation in host apps that already open the broadcast URL themselves — mitigated by the documented opt-outs, a README upgrade callout, and a CHANGELOG warning, but on by default. This includes **foreground taps**: a push banner tapped while the app is active now navigates immediately where it previously only broadcast. Secondary risks: unexpected browser launches from inbox taps with unclaimed `http(s)` action URLs (intended per ticket), and threading (opens are dispatched to the main actor via `Task { @MainActor in }`). Server-supplied URLs are validated before the SDK opens them (blocked schemes are still broadcast for host visibility, never opened). No network, persistence, or event-payload changes.

## Observability

No SDK crash telemetry; regressions would surface through customer/Zendesk reports and `os.log` output (searchable in Console.app, plus the debug log overlay during integration). Both open paths log success/failure, and refused schemes and disabled-flag skips are logged explicitly.

## Rollback

Runtime mitigations exist without a release: hosts can set `automaticallyOpensPushDeepLinks = false` / `automaticallyOpensInboxDeepLinks = false` and/or pass `onMessageTap` to suppress SDK-initiated navigation. A full rollback otherwise requires a follow-up SDK release and host-app adoption.

### DIFF
--- a/Sources/Helpers/Protocols/ATTNURLOpening.swift
+++ b/Sources/Helpers/Protocols/ATTNURLOpening.swift
@@ -1,0 +1,22 @@
+//
+//  ATTNURLOpening.swift
+//  attentive-ios-sdk-framework
+//
+//  Created by Umair Sharif on 7/24/26.
+//
+
+import UIKit
+
+/// Abstraction over `UIApplication`'s URL opening so deep-link handling can be unit tested.
+protocol ATTNURLOpening {
+    func open(_ url: URL, options: [UIApplication.OpenExternalURLOptionsKey: Any], completionHandler: ((Bool) -> Void)?)
+}
+
+/// Default `ATTNURLOpening` backed by `UIApplication.shared`.
+struct ATTNApplicationURLOpener: ATTNURLOpening {
+    func open(_ url: URL, options: [UIApplication.OpenExternalURLOptionsKey: Any], completionHandler: ((Bool) -> Void)?) {
+        Task { @MainActor in
+            UIApplication.shared.open(url, options: options, completionHandler: completionHandler)
+        }
+    }
+}
--- a/Sources/Helpers/Protocols/ATTNURLOpening.swift
+++ b/Sources/Helpers/Protocols/ATTNURLOpening.swift
@@ -23,20 +23,37 @@ struct ATTNApplicationURLOpener: ATTNURLOpening {
 
 extension URL {
     /// Schemes the SDK refuses to hand to `UIApplication.open` for server-driven deeplinks —
-    /// none of these are legitimate campaign destinations, and `javascript:`/`data:` are
-    /// script-injection vectors if a message payload is ever misconfigured or attacker-shaped.
-    private static let blockedDeepLinkSchemes: Set<String> = ["javascript", "file", "data", "about", "vbscript"]
+    /// none of these are legitimate campaign destinations. `javascript:`/`data:` are
+    /// script-injection vectors if a message payload is ever misconfigured or attacker-shaped;
+    /// the rest trigger system actions or prompts (dialer, SMS/mail composer, FaceTime call,
+    /// App Store / mobileconfig install) — an escalation a push tap must never cause without
+    /// host consent. Blocked URLs are still broadcast; hosts decide for themselves.
+    private static let blockedDeepLinkSchemes: Set<String> = [
+        // Scriptable / local-content schemes
+        "javascript", "file", "data", "about", "vbscript",
+        // Privileged system-action schemes
+        "tel", "telprompt", "sms", "mailto", "facetime", "facetime-audio",
+        "itms", "itms-apps", "itms-services",
+    ]
+
+    /// True when `scheme` is a shape the SDK will hand to `UIApplication.open`: non-empty,
+    /// scheme-shaped, and not on the blocked list. The shape check is deliberately looser than
+    /// RFC 3986 (which requires a leading letter and forbids `_`): underscores and leading
+    /// digits are registrable in Info.plist and the lenient pre-iOS 17 `URL(string:)` parses
+    /// them, so schemes like `my_app` or `1password` must not be dropped on iOS 15/16.
+    static func attnIsOpenableDeepLinkScheme(_ scheme: String?) -> Bool {
+        guard let scheme = scheme?.lowercased(),
+              scheme.range(of: "^[a-z0-9][a-z0-9_+.-]*$", options: .regularExpression) != nil else {
+            return false
+        }
+        return !blockedDeepLinkSchemes.contains(scheme)
+    }
 
     /// True when the URL is safe for the SDK to open as a deeplink. `URL(string:)` alone is too
     /// permissive for server-supplied strings: it accepts empty schemes (`"://foo"` parses with
-    /// `scheme == ""`) and scriptable schemes (`javascript:…`). Requires an RFC 3986-shaped
-    /// scheme (letter, then letters/digits/`+`/`-`/`.`) that is not on the blocked list;
-    /// `http(s)` and app custom schemes all pass.
+    /// `scheme == ""`), scriptable schemes (`javascript:…`), and privileged system-action
+    /// schemes (`tel:…`). `http(s)` and app custom schemes all pass.
     var attnIsOpenableDeepLink: Bool {
-        guard let scheme = scheme?.lowercased(),
-              scheme.range(of: "^[a-z][a-z0-9+.-]*$", options: .regularExpression) != nil else {
-            return false
-        }
-        return !Self.blockedDeepLinkSchemes.contains(scheme)
+        Self.attnIsOpenableDeepLinkScheme(scheme)
     }
 }
--- a/Sources/Helpers/Protocols/ATTNURLOpening.swift
+++ b/Sources/Helpers/Protocols/ATTNURLOpening.swift
@@ -20,3 +20,23 @@ struct ATTNApplicationURLOpener: ATTNURLOpening {
         }
     }
 }
+
+extension URL {
+    /// Schemes the SDK refuses to hand to `UIApplication.open` for server-driven deeplinks —
+    /// none of these are legitimate campaign destinations, and `javascript:`/`data:` are
+    /// script-injection vectors if a message payload is ever misconfigured or attacker-shaped.
+    private static let blockedDeepLinkSchemes: Set<String> = ["javascript", "file", "data", "about", "vbscript"]
+
+    /// True when the URL is safe for the SDK to open as a deeplink. `URL(string:)` alone is too
+    /// permissive for server-supplied strings: it accepts empty schemes (`"://foo"` parses with
+    /// `scheme == ""`) and scriptable schemes (`javascript:…`). Requires an RFC 3986-shaped
+    /// scheme (letter, then letters/digits/`+`/`-`/`.`) that is not on the blocked list;
+    /// `http(s)` and app custom schemes all pass.
+    var attnIsOpenableDeepLink: Bool {
+        guard let scheme = scheme?.lowercased(),
+              scheme.range(of: "^[a-z][a-z0-9+.-]*$", options: .regularExpression) != nil else {
+            return false
+        }
+        return !Self.blockedDeepLinkSchemes.contains(scheme)
+    }
+}
--- a/Sources/Inbox/InboxViewModel.swift
+++ b/Sources/Inbox/InboxViewModel.swift
@@ -97,11 +97,14 @@ class InboxViewModel: ObservableObject {
         }
     }
 
-    /// Called from `InboxView` when the user taps a row. In order: fires the click-tracking
-    /// POST + read flip, broadcasts `ATTNSDKInboxMessageTapped` (userInfo carries the
-    /// actionURL), then routes the tap — to the host's `onMessageTap` handler when one was
-    /// provided, otherwise by opening `actionURL` via `UIApplication`. Unclaimed http(s) links
-    /// fall back to the browser; a custom scheme no app claims is logged and dropped.
+    /// Called from `InboxView` when the user taps a row. Dispatches the click-tracking POST +
+    /// read flip to the manager (async — navigation must not wait on the network, matching the
+    /// Android SDK), broadcasts `ATTNSDKInboxMessageTapped` (userInfo carries the actionURL),
+    /// then routes the tap — to the host's `onMessageTap` handler when one was provided,
+    /// otherwise by opening `actionURL` via `UIApplication`. Unclaimed http(s) links fall back
+    /// to the browser; a custom scheme no app claims is logged and dropped. Note the tracking
+    /// runs concurrently with routing: an `onMessageTap` handler that reads inbox state
+    /// synchronously may still observe the message as unread.
     func click(_ message: Message) {
         Task {
             await inboxManager.markClicked(message.id)
@@ -124,6 +127,12 @@ class InboxViewModel: ObservableObject {
         }
 
         guard let actionURL = message.actionURL else { return }
+        // Server-supplied string: refuse empty-scheme and scriptable (javascript:/file:/data:)
+        // URLs. The broadcast above still carries the raw URL — hosts decide for themselves.
+        guard actionURL.attnIsOpenableDeepLink else {
+            Loggers.network.error("Refusing to open inbox action URL with unsupported scheme: \(actionURL, privacy: .public)")
+            return
+        }
         urlOpener.open(actionURL, options: [:]) { success in
             if !success {
                 Loggers.network.error("No app claimed inbox action URL: \(actionURL, privacy: .public)")
--- a/Sources/Inbox/InboxViewModel.swift
+++ b/Sources/Inbox/InboxViewModel.swift
@@ -6,6 +6,7 @@
 //
 
 import Foundation
+import UIKit
 
 @MainActor
 class InboxViewModel: ObservableObject {
@@ -28,12 +29,21 @@ class InboxViewModel: ObservableObject {
     let style: InboxStyle
 
     private let inboxManager: InboxManager
+    private let onTap: ((Message) -> Void)?
+    private let urlOpener: ATTNURLOpening
     private var stateStreamTask: Task<Void, Never>?
     private var loadingMoreStreamTask: Task<Void, Never>?
 
-    init(inboxManager: InboxManager, style: InboxStyle) {
+    init(
+        inboxManager: InboxManager,
+        style: InboxStyle,
+        onTap: ((Message) -> Void)? = nil,
+        urlOpener: ATTNURLOpening = ATTNApplicationURLOpener()
+    ) {
         self.inboxManager = inboxManager
         self.style = style
+        self.onTap = onTap
+        self.urlOpener = urlOpener
         state = .loading
         stateStreamTask = Task { [weak self] in
             guard let stream = await self?.inboxManager.stateStream else { return }
@@ -87,15 +97,17 @@ class InboxViewModel: ObservableObject {
         }
     }
 
-    /// Called from `InboxView` when the user taps a row. Fires the click-tracking POST + flips
-    /// the row's read state, and broadcasts `ATTNSDKInboxMessageTapped` so host apps can route
-    /// to `actionURL`. The SDK intentionally does not open the URL itself.
-    func click(id: Message.ID, actionURL: URL?) {
+    /// Called from `InboxView` when the user taps a row. In order: fires the click-tracking
+    /// POST + read flip, broadcasts `ATTNSDKInboxMessageTapped` (userInfo carries the
+    /// actionURL), then routes the tap — to the host's `onMessageTap` handler when one was
+    /// provided, otherwise by opening `actionURL` via `UIApplication`. Unclaimed http(s) links
+    /// fall back to the browser; a custom scheme no app claims is logged and dropped.
+    func click(_ message: Message) {
         Task {
-            await inboxManager.markClicked(id)
+            await inboxManager.markClicked(message.id)
         }
-        var userInfo: [AnyHashable: Any] = ["attentiveInboxMessageId": id]
-        if let actionURL {
+        var userInfo: [AnyHashable: Any] = ["attentiveInboxMessageId": message.id]
+        if let actionURL = message.actionURL {
             userInfo["attentiveInboxActionUrl"] = actionURL
         }
         NotificationCenter.default.post(
@@ -103,6 +115,20 @@ class InboxViewModel: ObservableObject {
             object: nil,
             userInfo: userInfo
         )
+
+        // Host-provided handler replaces the SDK's routing entirely (matches the Android
+        // SDK's onMessageClick override) — it hears every tap, even URL-less ones.
+        if let onTap {
+            onTap(message)
+            return
+        }
+
+        guard let actionURL = message.actionURL else { return }
+        urlOpener.open(actionURL, options: [:]) { success in
+            if !success {
+                Loggers.network.error("No app claimed inbox action URL: \(actionURL, privacy: .public)")
+            }
+        }
     }
 }
 
--- a/Sources/Inbox/InboxViewModel.swift
+++ b/Sources/Inbox/InboxViewModel.swift
@@ -31,6 +31,10 @@ class InboxViewModel: ObservableObject {
     private let inboxManager: InboxManager
     private let onTap: ((Message) -> Void)?
     private let urlOpener: ATTNURLOpening
+    /// Read at tap time (not captured at init) so hosts can toggle
+    /// `ATTNSDK.automaticallyOpensInboxDeepLinks` after the inbox view is created,
+    /// matching the push flag's read-on-tap semantics.
+    private let shouldOpenDeepLink: () -> Bool
     private var stateStreamTask: Task<Void, Never>?
     private var loadingMoreStreamTask: Task<Void, Never>?
 
@@ -38,12 +42,14 @@ class InboxViewModel: ObservableObject {
         inboxManager: InboxManager,
         style: InboxStyle,
         onTap: ((Message) -> Void)? = nil,
-        urlOpener: ATTNURLOpening = ATTNApplicationURLOpener()
+        urlOpener: ATTNURLOpening = ATTNApplicationURLOpener(),
+        shouldOpenDeepLink: @escaping () -> Bool = { true }
     ) {
         self.inboxManager = inboxManager
         self.style = style
         self.onTap = onTap
         self.urlOpener = urlOpener
+        self.shouldOpenDeepLink = shouldOpenDeepLink
         state = .loading
         stateStreamTask = Task { [weak self] in
             guard let stream = await self?.inboxManager.stateStream else { return }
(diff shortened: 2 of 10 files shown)


usharif-attentive:
@codex review

usharif-attentive:
@attentive-claude review

Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -930,6 +917,65 @@ public final class ATTNSDK: NSObject {
...
+    private func openDeepLink(_ url: URL) {
+        guard automaticallyOpensPushDeepLinks else {
+            Loggers.network.debug("Automatic deep link opening is disabled; host app is expected to handle URL: \(url, privacy: .public)")
+            return
+        }
+
+        let isWebLink = ["http", "https"].contains(url.scheme?.lowercased() ?? "")
+        // Web links must resolve as universal links into an app; never bounce the user to the browser.
+        let options: [UIApplication.OpenExternalURLOptionsKey: Any] = isWebLink ? [.universalLinksOnly: true] : [:]
+        urlOpener.open(url, options: options) { success in
```
adelag-attentive, line 1001: **URL validation upstream of this line is too permissive for a server-driven deeplink.** In `normalizeAndBroadcast`, the guard is `trimmedURL.scheme != nil`. `URL(string:)` accepts several shapes that shouldn't be handed to `UIApplication.open`:

- `"://foo"` → parses with `scheme = ""` (empty string, not nil). Passes the guard.
- `"javascript:alert(1)"` → parses with `scheme = "javascript"`. Modern iOS refuses `javascript:` from `UIApplication.open` most of the time, but relying on that is defense-through-obscurity.
- `"file:///tmp/x"` → parses with `scheme = "file"`.

`attentive_open_action_url` comes from server-supplied `attentiveCallbackData` and passes through `escapeJSONDictionary` untouched. If the server is ever misconfigured or a marketer pastes the wrong thing into a campaign, the SDK will happily hand any URL scheme to `UIApplication.open`. Worth checking that the Android SDK's equivalent path has the same allowlist stance.

Suggest tightening to: scheme is non-empty AND matches `[a-zA-Z][a-zA-Z0-9+.\-]*` AND is in an explicit allowlist (`http`, `https`, or app custom schemes — explicitly reject `javascript`, `file`, `data`, `about`, `vbscript`). At minimum reject empty-scheme URLs. A regression test for `"://foo"` and `"javascript:..."` on `normalizeAndBroadcast` would catch this cheaply.

Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -72,6 +72,20 @@ public final class ATTNSDK: NSObject {
     /// Determinates if fatigue rules evaluation will be skipped for Creative. Default value is false.
     @objc public var skipFatigueOnCreative: Bool = false
 
+    /// Determines if the SDK opens the deep link from a tapped push notification
+    /// (`attentive_open_action_url`) on the host app's behalf. Default value is true.
```
adelag-attentive, line 76: **Docstring gap on the foreground-tap behavior change.** `handleForegroundPush` (unchanged upstream, but now calls `normalizeAndBroadcast` which navigates) means: a user reading their inbox in-app and getting a push banner about "50% off" that they tap will now be pulled to a product screen. Before this PR, the SDK just broadcast the URL for the foreground case.

That's likely the correct Android-parity behavior, but this docstring only mentions the opt-out for hosts that navigate via broadcasts. Worth explicitly naming the foreground-active-tap ergonomic change too — hosts that today rely on "push arrives while in-app, user taps banner, we just log it" will see live navigation start happening on the next release. `automaticallyOpensPushDeepLinks = false` covers this case, but the docstring should call it out so hosts don't get surprised. Also worth a line under **Risk** in the PR body.

Tests/Doubles/Spies/ATTNURLOpenerSpy.swift
```diff
@@ -0,0 +1,24 @@
...
+    private(set) var openedURLs: [URL] = []
+    private(set) var lastOptions: [UIApplication.OpenExternalURLOptionsKey: Any] = [:]
+
+    var openResult = true
+
+    func open(_ url: URL, options: [UIApplication.OpenExternalURLOptionsKey: Any], completionHandler: ((Bool) -> Void)?) {
+        openWasCalled = true
+        openedURLs.append(url)
+        lastOptions = options
+        completionHandler?(openResult)
```
adelag-attentive: The spy invokes `completionHandler?(openResult)` synchronously inside `open`. The real `ATTNApplicationURLOpener` dispatches through `Task { @MainActor in ... }` and then gets its completion off `UIApplication.shared.open`'s async callback — the completion runs on a later runloop turn.

Any test that asserts state *right after* `viewModel.click(...)` returns will pass with the spy (completion has fired synchronously, `openedURLs`/`openWasCalled` are set) but the same production code path would race in the field. Not a bug right now — the callback here only logs — but it's a foot-gun for anyone who later moves logic into the completion handler and writes a test that only exercises the spy.

Two cheap options: (1) hop the spy's completion onto the main queue with `DispatchQueue.main.async` so the spy's timing matches the real one, or (2) add a `synchronous: Bool = false` flag defaulting to async and let tests opt in. (1) is simpler and would probably shake out any latent ordering assumptions in the existing tests.

Sources/Inbox/InboxViewModel.swift
```diff
@@ -87,22 +97,38 @@ class InboxViewModel: ObservableObject {
...
+
+        // Host-provided handler replaces the SDK's routing entirely (matches the Android
+        // SDK's onMessageClick override) — it hears every tap, even URL-less ones.
+        if let onTap {
+            onTap(message)
+            return
+        }
+
+        guard let actionURL = message.actionURL else { return }
+        urlOpener.open(actionURL, options: [:]) { success in
```
adelag-attentive, line 151: Nit / doc gap: there's no ergonomic way for a host to say "broadcast + track, but don't navigate." `onMessageTap: { _ in }` works but is non-obvious as an opt-out pattern (the empty closure reads like a mistake).

Compare with the push side, which exposes `automaticallyOpensPushDeepLinks: Bool = true` as an explicit toggle. Inbox could do the same — either a matching `automaticallyOpensInboxDeepLinks: Bool` property, or a documented `onMessageTap: { _ in }` sentinel in the docstring ("Pass an empty closure to suppress SDK navigation entirely while keeping click tracking + broadcast").

Not a merge blocker; something to align on with the inbox alpha hosts (JTV/Aero/JCP) before GA.

Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -72,6 +72,20 @@ public final class ATTNSDK: NSObject {
...
+    ///
+    /// Custom-scheme links are handed to `UIApplication`, which routes them back into the host
+    /// app's URL handlers. Http(s) links are opened as universal links only — if no app has
+    /// registered for them they are never sent to the browser; the URL stays available through
+    /// the `ATTNSDKDeepLinkReceived` broadcast and `consumeDeepLink()`.
+    ///
+    /// Set to false if the host app navigates on its own via the `ATTNSDKDeepLinkReceived`
+    /// broadcast or `consumeDeepLink()`, to avoid handling the same URL twice.
+    @objc public var automaticallyOpensPushDeepLinks: Bool = true
```
adelag-attentive, line 91: Nit: `automaticallyOpensPushDeepLinks` is `@objc public var` with no synchronization guarantee. A host toggling it from a background thread while a push tap is being handled on the main queue races with the read at line ~961. In practice hosts set this once at init and never touch it — but if you want to lock it down, `@MainActor`-annotate the setter or add a note in the docstring saying it must be set on the main thread.

adelag-attentive (commented):
Nice extraction of `ATTNURLOpening` — the injection point makes the push + inbox paths cleanly testable, and the Android-parity behavior is correct in principle. Five things worth addressing before merge — mostly around input validation on server-supplied deeplinks, one behavior-change framing gap, and a test-fidelity gap in the URL opener spy. Detail inline.

Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -72,6 +72,26 @@ public final class ATTNSDK: NSObject {
...
+    ///
+    /// This applies to foreground taps too: when a push banner arrives while the app is active
+    /// and the user taps it, the SDK navigates immediately — previously it only broadcast the
+    /// URL. Hosts that treat in-app banner taps as log-only should opt out.
+    ///
+    /// Set to false if the host app navigates on its own via the `ATTNSDKDeepLinkReceived`
+    /// broadcast or `consumeDeepLink()`, to avoid handling the same URL twice. Set it once
+    /// during SDK setup on the main thread — it is read on the main thread when a tap is
+    /// handled, and mutation from other threads is not synchronized.
+    @objc public var automaticallyOpensPushDeepLinks: Bool = true
```
adelag-attentive, line 91: **Breaking behavior change on SDK upgrade: hosts already handling `.ATTNSDKDeepLinkReceived` will double-navigate.** Default `= true` means every existing integration that observes the broadcast and opens the URL itself now sees the SDK also call `UIApplication.shared.open` on the same URL. For custom schemes: `application(_:open:options:)` fires twice → target screen pushed twice, checkout state wiped, back-nav broken. Cold-launch is worse — `handlePushOpen` races the host's documented `launchOptions[.remoteNotification]` handling on the same URL.

The repo's backward-compat guardrail (README + CLAUDE.md guidance) says: "Never change the type or behavior of an emitted callback path." A default-on flag flip qualifies.

**Suggested fix:** default to `false` for at least one release cycle so existing integrations aren't broken by an SDK bump alone. Hosts that want the new behavior can opt in with a one-line change. After enough adoption, flip the default in a major version.

Sources/Inbox/InboxViewModel.swift
```diff
@@ -87,22 +97,47 @@ class InboxViewModel: ObservableObject {
...
+        guard actionURL.attnIsOpenableDeepLink else {
+            Loggers.network.error("Refusing to open inbox action URL with unsupported scheme: \(actionURL, privacy: .public)")
+            return
+        }
+        urlOpener.open(actionURL, options: [:]) { success in
+            if !success {
+                Loggers.network.error("No app claimed inbox action URL: \(actionURL, privacy: .public)")
+            }
+        }
     }
```
adelag-attentive, line 156: **Same class of breaking change as `automaticallyOpensPushDeepLinks`.** The old inline comment at `InboxView.swift` (lines ~31-34, deleted in this PR) documented the contract:

> "Fires click tracking + local read flip, and broadcasts `ATTNSDKInboxMessageTapped` (userInfo carries the actionURL). The SDK does not open the URL itself — host apps route it."

That's the contract every existing integrator was building against. This PR inverts it: post-tap, the SDK broadcasts AND opens by default. Any host that already observes `.ATTNSDKInboxMessageTapped` and calls `UIApplication.shared.open(url)` in response now double-navigates on the same tap.

**The opt-out is non-obvious.** `onMessageTap: { _ in }` (empty closure = no-op) reads like a mistake, not "please don't open." A host reviewing their integration wouldn't write it that way without documentation pointing them at it.

**Suggested fix:** default `onMessageTap` to `nil` meaning "broadcast only, don't open." Hosts that want SDK-driven navigation opt in explicitly (e.g., pass `onMessageTap: .openAutomatically` or a matching `automaticallyOpensInboxDeepLinks` flag on ATTNSDK to mirror the push side). Related to my earlier nit on line 136 about the opt-out API shape being inconsistent with push.

Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -930,6 +924,68 @@ public final class ATTNSDK: NSObject {
...
+        if let trimmedURL = URL(string: trimmed), trimmedURL.attnIsOpenableDeepLink {
+            candidateURL = trimmedURL
+        }
+
+        guard let validURL = candidateURL else {
+            Loggers.network.error("Failed to parse deep link URL from string: '\(trimmed, privacy: .public)' - Visitor ID: \(self.userIdentity.visitorId, privacy: .public)")
+            return
+        }
+
+        pendingURL = validURL
```
adelag-attentive, line 970: **`normalizeAndBroadcast` gates BOTH the broadcast AND the `pendingURL` write on `attnIsOpenableDeepLink`, silently dropping URLs whose scheme is blocked or non-RFC.**

Pre-PR: any URL that parsed with a non-nil scheme was stored in `pendingURL` and broadcast. Hosts could observe `.ATTNSDKDeepLinkReceived` for logging, security audit, or their own scheme handling — even for shapes the SDK considers unsafe.

Post-PR: `javascript:`, `file:`, `data:`, `about:`, `vbscript:`, `://foo` (empty scheme), and custom schemes with `_` or leading digits all hit the `guard let validURL = candidateURL else { return }` and never make it to `pendingURL`, `NotificationCenter.post`, or `consumeDeepLink()`. Even setting `automaticallyOpensPushDeepLinks = false` does not restore visibility — the validation runs before the flag is checked.

**The same PR uses the correct pattern in the inbox path** (`Sources/Inbox/InboxViewModel.swift:~130`): broadcast unconditionally, gate only the `urlOpener.open` call, with an explicit comment "The broadcast above still carries the raw URL — hosts decide for themselves." The push path should mirror it.

**Suggested fix:**

```swift
// Broadcast for any URL that parses (host-visibility is the goal here)
guard let trimmedURL = URL(string: trimmed), trimmedURL.scheme?.isEmpty == false else { ... }
pendingURL = trimmedURL
NotificationCenter.default.post(...)

// Gate the SDK-initiated open on both the safety check and the flag
if trimmedURL.attnIsOpenableDeepLink {
    openDeepLink(trimmedURL)
}
```

This preserves the security defense on the open path while restoring host visibility for logging/audit consumers.

Sources/Helpers/Protocols/ATTNURLOpening.swift
```diff
@@ -0,0 +1,42 @@
...
+            UIApplication.shared.open(url, options: options, completionHandler: completionHandler)
+        }
+    }
+}
+
+extension URL {
+    /// Schemes the SDK refuses to hand to `UIApplication.open` for server-driven deeplinks —
+    /// none of these are legitimate campaign destinations, and `javascript:`/`data:` are
+    /// script-injection vectors if a message payload is ever misconfigured or attacker-shaped.
+    private static let blockedDeepLinkSchemes: Set<String> = ["javascript", "file", "data", "about", "vbscript"]
```
adelag-attentive: **Blocklist misses privileged-action schemes.** The current list — `["javascript", "file", "data", "about", "vbscript"]` — blocks script-injection schemes but not schemes that trigger *system prompts* (dialer, SMS composer, App Store, FaceTime, etc.). All of these currently pass `attnIsOpenableDeepLink` and reach `UIApplication.shared.open`:

- `tel:+15551234` → system dial prompt
- `sms:+15551234` → Messages composer
- `mailto:foo@bar` → Mail composer
- `facetime://+15551234` / `facetime-audio://...` → FaceTime call prompt
- `itms-apps://...` / `itms-services://...` → App Store / mobileconfig install prompt

**Failure scenario:** merchant Ops accidentally shapes a campaign as `attentive_open_action_url = "tel:+15551234"`. User taps the push from lock screen → SDK calls `UIApplication.shared.open("tel:...")` → iOS shows a call-confirmation prompt. That's an escalation from "post a NotificationCenter broadcast" to "trigger a system action from a background push tap without host consent."

The file's own comment argues *"none of these are legitimate campaign destinations"* about the current blocklist. The same reasoning applies to `tel`/`sms`/`mailto`/`facetime`/`itms-*` — none of them are legitimate deeplink campaign destinations either, and if any of them ever become legitimate a host can register a custom scheme instead.

**Suggested fix:** either expand the blocklist to include at minimum `tel`, `sms`, `mailto`, `facetime`, `facetime-audio`, `itms`, `itms-apps`, `itms-services`, `st`, `iBooks`, `music`, `podcast`, or (safer, harder to bypass) invert to an allowlist keyed by scheme category (custom app schemes + http/https + a merchant-approved list).

Sources/Helpers/Protocols/ATTNURLOpening.swift
```diff
@@ -0,0 +1,42 @@
...
+
+    /// True when the URL is safe for the SDK to open as a deeplink. `URL(string:)` alone is too
+    /// permissive for server-supplied strings: it accepts empty schemes (`"://foo"` parses with
+    /// `scheme == ""`) and scriptable schemes (`javascript:…`). Requires an RFC 3986-shaped
+    /// scheme (letter, then letters/digits/`+`/`-`/`.`) that is not on the blocked list;
+    /// `http(s)` and app custom schemes all pass.
+    var attnIsOpenableDeepLink: Bool {
+        guard let scheme = scheme?.lowercased(),
+              scheme.range(of: "^[a-z][a-z0-9+.-]*$", options: .regularExpression) != nil else {
+            return false
```
adelag-attentive, line 47: **Scheme regex `^[a-z][a-z0-9+.-]*$` strictly follows RFC 3986 but silently rejects real-world custom schemes that iOS accepts** — specifically schemes with underscores or leading digits:

- `my_app://cart` (underscore in the middle)
- `1password://vault/item`
- `firebase_dynamiclinks://...`
- `com_myapp://...`

`URL(string:)` on iOS 15/16 (which this SDK still supports per `Package.swift: platforms: [.iOS(.v15)]`) is lenient and accepts these. Pre-PR the `scheme != nil` check let them through; post-PR the regex silently rejects them.

**Failure scenarios:**
- **Inbox:** row with `actionURL = "my_app://sale"` — user taps, sees no visible feedback (SDK logs "unsupported scheme" and drops the tap). Dead tap UX.
- **Push:** same URL in `attentive_open_action_url` — silent drop, no broadcast, no `pendingURL`, no `consumeDeepLink()` return. The campaign never reaches the host.

Note: on iOS 17+ `URL(string:)` itself became stricter and rejects these, so on newer OS versions this isn't a regression. But the SDK's minimum-supported iOS is 15, so this is a real regression for merchants whose users are on iOS 15/16.

**Suggested fix:** either drop the regex entirely and rely on the (expanded — see my other comment) blocklist, or relax the regex to allow `_` in the tail:

```swift
private static let schemeRegex = "^[a-z][a-z0-9_+.-]*$"
```

`URL(string:)` already refuses truly malformed schemes; the regex is doing redundant work in most cases and rejecting valid ones in the edge case.

Sources/Public/SDK/ATTNSDK.swift
usharif-attentive: Good catch — agreed this regressed host visibility, and the inconsistency with the inbox path made it worse. Fixed in dabc152: `normalizeAndBroadcast` now stores + broadcasts any URL that parses with a non-empty scheme (the pre-PR contract, minus the never-usable `"://foo"` empty-scheme shape), and only the SDK-initiated open is gated on `attnIsOpenableDeepLink` — matching the inbox path exactly. `automaticallyOpensPushDeepLinks = false` now behaves as documented for every parseable URL, and the scriptable-schemes test pins the new contract (broadcast + consumable, never opened).

Sources/Helpers/Protocols/ATTNURLOpening.swift
usharif-attentive: Agreed — the "none of these are legitimate campaign destinations" reasoning applies just as much to system-action schemes. Fixed in dabc152: the blocklist now also covers `tel`, `telprompt` (legacy no-confirmation dialer variant), `sms`, `mailto`, `facetime`, `facetime-audio`, `itms`, `itms-apps`, `itms-services`, with a regression test.

I stayed with a blocklist rather than inverting to an allowlist: the SDK can't enumerate the host's registered custom schemes or a merchant-approved list without new config surface, and with the broadcast now un-gated (per your other comment), the worst case for an unlisted scheme is "SDK doesn't auto-open; host still sees the broadcast." Happy to revisit an allowlist if we add per-merchant config.

Sources/Helpers/Protocols/ATTNURLOpening.swift
usharif-attentive: Fixed in dabc152: relaxed to `^[a-z0-9][a-z0-9_+.-]*$` — underscores *and* leading digits, since your `1password://` example was rejected by the leading-`[a-z]` anchor too, not just the `_` gap. Kept the regex rather than dropping it entirely — it's what rejects the empty-scheme `"://foo"` shape on the open path — but it no longer rejects anything the lenient iOS 15/16 parser produces for registrable schemes.

One testability note: the iOS 17+ parser used by the test host refuses to construct `URL("my_app://…")` at all, so the check is now exposed as `URL.attnIsOpenableDeepLinkScheme(_:)` and the regression tests exercise the scheme matrix directly (`my_app`, `1password`, `firebase_dynamiclinks`, `web+shop`, …). Severity on the push path also dropped with the broadcast fix — a non-RFC scheme on iOS 15/16 now still broadcasts + stores; only the auto-open depends on the shape check.

Sources/Public/SDK/ATTNSDK.swift
usharif-attentive: Fair challenge — I went and checked what Android actually does before deciding, and it changed my read on the risk.

**On Android, SDK-driven navigation is not just the default — it is the only behavior.** `AttentivePush.buildLaunchIntent` bakes the deeplink into the notification tap as `Intent(ACTION_VIEW, deepLink.toUri())`; there is no flag, no opt-out, and crucially no broadcast/callback alternative at all. No Android host has ever navigated off an SDK signal, so the double-navigation hazard is purely an iOS artifact of having shipped the broadcast-only contract first.

**The at-risk population is also smaller than "every existing integration":** `ATTNSDKDeepLinkReceived` only shipped in the push beta merge (Oct 2025). The exposed set is "hosts that adopted iOS push in the last ~10 months AND wired navigation off the broadcast" — recent, small, and probably enumerable by CS.

Given that, I want to keep the default `true` (a `false` default means Android/iOS parity never actually materializes for the long tail), and mitigate instead:

1. Loud CHANGELOG + README migration callout: if you navigate off `ATTNSDKDeepLinkReceived` or `consumeDeepLink()`, set `automaticallyOpensPushDeepLinks = false` before upgrading.
2. Have CS ping the known push-beta adopters directly before the release goes out.

If you feel strongly that this still violates the backward-compat guardrail, happy to take it to the team — but "match Android, which has always auto-opened" plus a targeted heads-up to a known-small cohort feels like the right trade.

Sources/Inbox/InboxViewModel.swift
usharif-attentive: Checked Android for precedent here too: `AttentiveInbox`'s `onMessageClick` defaults to `null`, and the null case falls through to an SDK handler that marks read, tracks the click, then `startActivity(Intent(ACTION_VIEW, url))` — unscoped, so browser fallback included. A host-provided `onMessageClick` replaces the default entirely. In other words, Android inbox is already exactly the shape this PR gives iOS: SDK-opens by default, full override via callback.

On the contract inversion: agreed the deleted comment documented broadcast-only, but inbox is pre-GA with three alpha hosts (JTV/Aero/JCP) — this is the cheapest moment to converge on the contract we actually want (Android's), rather than freezing a pre-GA comment into API. I'll confirm with the alpha hosts that none of them currently navigate off `ATTNSDKInboxMessageTapped` before merge.

On the opt-out ergonomics: agreed the empty closure is too magical. I'll add a mirrored `automaticallyOpensInboxDeepLinks` flag on `ATTNSDK` (default `true`) so both surfaces opt out the same way — tracking + broadcast still fire, only the SDK-initiated open is suppressed. That also resolves your round-1 nit on this line. Will push that in a follow-up commit on this branch.

usharif-attentive: `automaticallyOpensInboxDeepLinks` landed in a0f56c2 — default `true`, read at tap time, suppresses only the SDK-initiated open (click tracking + `ATTNSDKInboxMessageTapped` still fire), and the `inboxView()` docs now point at the flag instead of the empty-closure trick. Still owe you the alpha-host confirmation before merge.

usharif-attentive:
@codex review

usharif-attentive:
@attentive-claude review

adelag-attentive (approved):

PR #286, opened 2026-07-28
MSDK-441: fix v2 event body encoding and restore pd deeplink

adelag-attentive:
## PR Type
- [x] Bugfix

[MSDK-441](https://attentivemobile.atlassian.net/browse/MSDK-441) — Carters (3732) bug in iOS SDK 2.0.15; related [SEV-4899](https://attentivemobile.atlassian.net/browse/SEV-4899).

## What's new?

Two independent bugs surfaced by Carters after upgrading to 2.0.15 on 2026-07-15.

### 1. v2 event body truncation on `&`/`=`/`+`/`#`

`ATTNAPI.sendNewEvent` posts JSON as `application/x-www-form-urlencoded` with body `d=<json>`, but percent-encoded the JSON with `.urlQueryAllowed` — a set that leaves the form sub-delims unescaped. Any of those characters in an event payload (a product name like `"Grab & Go"`, `"Bodysuit & Pant Set"`, etc.) splits the form body: the servlet parser terminates the `d` value at the first `&`, Jackson throws `JsonProcessingException` on the truncated JSON, AEH (`InternalEventsUnifiedController.java:47-79`) returns `422 Unprocessable Entity` and increments `recordUnifiedApiJsonParseError("mobile")`. Event dropped.

**Fix:** introduce a file-local `CharacterSet.attnFormEncodedAllowed` = `.urlQueryAllowed` minus `&=+#`, and apply it in the body-encoding call.

### 2. v2 endpoint dropped `pd` deeplink

`sendNewEventInternal` built its `ATTNEventRequest` without a `deeplink`, so `ProductView` and `AddToCart` sent through the v2 path lost the deeplink even though the URL builder already emitted `pd` when set. Threaded `deeplink` from the public event through `sendLegacyEventAsV2` → `sendAddToCartEvent` / `sendProductViewEvent` → `sendNewEventInternal` → `eventRequest.deeplink`. All new params are defaulted (`String? = nil`) so the two-arg external callers (`ATTNEventTracker.swift:72, 75`) and existing tests compile unchanged.

### What is NOT changing
- Public API surface: `ATTNSDK.send(event:)`, `ATTNProductViewEvent`, `ATTNAddToCartEvent` — signatures identical.
- Legacy `/e` endpoint code path — untouched; already safe because it builds the query via `URLComponents.queryItems` which percent-encodes sub-delims in values.
- Push token, opt-in/out, app-events, identity paths.
- Content-Type, endpoint URL, retry behavior, callback contract.

## Testing

- **Unit tests added** (14 new, all pass):
  - `ATTNAPINewEventEncodingTests` — real `ATTNAPI` + `NSURLSessionMock`; inspects raw `httpBody` and parses via `URLComponents(percentEncodedQuery:)`. Covers `&`, `=`, `+`, `#`, all four combined, plus the three literal product names from the SEV report (`"Grab & Go Snacks"`, `"Ribbed Top & Shorts Set"`, `"Bodysuit & Pant Set"`). Non-regression case with safe characters and Content-Type header assertion.
  - `ATTNv2DeeplinkTests` — drives the full `sdk.send(event:)` path with `useV2Endpoint=true`; asserts `deeplink` on the captured `ATTNEventRequest`; end-to-end URL check verifies `pd=` appears on the outbound URL.

- **Full suite:** 216 tests, 0 failures. SwiftLint clean.

- **Not automated yet — recommended before release:** point a Bonni build at a scratch domain with Charles capture on, send an AddToCart with product name `"Grab & Go Snacks"`, confirm the outbound body is a single `d=<percent-encoded-json>` field with `%26` in place of `&`.

## Rollback / host-app impact
Zero. No public API changes; no dependency or config changes. Reverting this commit restores the previous behavior.

## Screenshots
n/a — networking change.

[MSDK-441]: https://attentivemobile.atlassian.net/browse/MSDK-441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SEV-4899]: https://attentivemobile.atlassian.net/browse/SEV-4899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/API/ATTNAPI.swift
+++ b/Sources/API/ATTNAPI.swift
@@ -510,9 +510,9 @@ extension ATTNAPI {
             let jsonString = String(data: jsonData, encoding: .utf8) ?? ""
 
             // Percent-encode the JSON for an `application/x-www-form-urlencoded` body.
-            // `.urlQueryAllowed` leaves sub-delims (`&`, `=`, `+`, `#`) unescaped, which
-            // truncates the `d` value at the first such character in the JSON (e.g. a
-            // product name like "Grab & Go" splits the body and drops the event).
+            // `.urlQueryAllowed` leaves sub-delims (`&`, `=`, `+`) unescaped, which
+            // truncates the `d` value at the first such character (e.g. a product
+            // name like "Grab & Go" splits the body and drops the event).
             guard let encodedJson = jsonString.addingPercentEncoding(withAllowedCharacters: .attnFormEncodedAllowed) else {
                 Loggers.network.error("Failed to URL-encode JSON payload")
                 callback?(nil, url, nil, ATTNError.badURL)
@@ -548,15 +548,4 @@ extension ATTNAPI {
         }
     }
 
-}
-
-private extension CharacterSet {
-    /// Percent-encoding allow-set for values inside an `application/x-www-form-urlencoded`
-    /// body: `.urlQueryAllowed` minus the sub-delims that separate form fields (`&`, `=`)
-    /// or have special decode semantics in a form body (`+`, `#`).
-    static let attnFormEncodedAllowed: CharacterSet = {
-        var set = CharacterSet.urlQueryAllowed
-        set.remove(charactersIn: "&=+#")
-        return set
-    }()
 }
--- a/Sources/API/ATTNAPI.swift
+++ b/Sources/API/ATTNAPI.swift
@@ -509,8 +509,11 @@ extension ATTNAPI {
             // Convert JSON to string for logging
             let jsonString = String(data: jsonData, encoding: .utf8) ?? ""
 
-            // URL-encode the JSON and wrap it in form data with key 'd'
-            guard let encodedJson = jsonString.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) else {
+            // Percent-encode the JSON for an `application/x-www-form-urlencoded` body.
+            // `.urlQueryAllowed` leaves sub-delims (`&`, `=`, `+`, `#`) unescaped, which
+            // truncates the `d` value at the first such character in the JSON (e.g. a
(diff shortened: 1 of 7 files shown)


Sources/API/ATTNAPI.swift
```diff
@@ -546,3 +549,14 @@ extension ATTNAPI {
     }
 
 }
+
```
usharif-attentive: The repo already has this exact concept in `ATTNEventURLProvider.formEncodedAllowedCharacters` (`.urlQueryAllowed` minus `+&=`, used by `setFormEncodedQuery`). This adds a second, file-private near-copy — two sets serving the same purpose in two files will drift (they already differ on `#`, which is a no-op removal since `#` is not in `.urlQueryAllowed`).

Suggest consolidating into a single internal `CharacterSet` extension used by both call sites — this is exactly the kind of subtle constant we want defined once. At minimum, a cross-reference comment in both places would help.

adelag-attentive: Good catch — I missed the existing one because it's private on `main` and I based my initial grep on `develop` where MSDK-354 hadn't landed yet.

Consolidated in `fc241dd`: moved to `Sources/Helpers/Extension/CharacterSet+Extension.swift` as `CharacterSet.attnFormEncodedAllowed`. Removed both the private one in `ATTNEventURLProvider` and the file-private one I introduced in `ATTNAPI`; both call sites now point at the shared symbol. Also merged latest `main` (picks up MSDK-442).

usharif-attentive (approved):

PR #291, opened 2026-08-04
MSDK-450: tear down WebView when a creative fails to render

adelag-attentive:
## Summary

- Reported by the Target iOS team: when a creative fails to render (fatigued, timed out, network error, WebContent crash), the SDK was leaving a full-screen invisible `WKWebView` pinned to the parent view. FullStory picked it up, VoiceOver could reach it, and every trigger burned a WebContent process + creative fetch even when nothing rendered — blocking their expansion beyond a handful of pages.
- Now every failure path routes through a single teardown primitive that detaches the delegate, removes the view from its parent, stops the load, clears scripts, and nils the provider's reference. `closeCreative` (successful dismissal) and the failure path share the same primitive, so cleanup stays in lockstep.
- Every async callback captures a monotonic `currentLaunchEpoch` and no-ops if a newer launch has started — this closes a multi-webview race where a stale timer from a failed launch #1 could tear down a healthy WebView belonging to launch #2.
- Covered failure modes: native 5s launching timeout, JS iframe-detection timeout, `didFail` / `didFailProvisionalNavigation`, `webViewWebContentProcessDidTerminate`, and `onRemovedFromWindow` mid-launch. Also fixed a latent race where the native timer could tear down a *successfully* rendered creative in the window between `.opened` firing and IMPRESSION arriving.

## Test plan

Automated:
- [x] `bundle exec fastlane ios unit_test` — 226 tests, 0 failures. 3 new integration tests exercise the native-timeout, JS-timeout, and racing-timeout paths.
- [x] `bundle exec fastlane ios lint` — clean.

Manual (already completed):
- [x] **Working case** — Triggered the `games` domain creative (renders successfully). Log shows `Found creative iframe, showing WebView.` → IMPRESSION → RESIZE_FRAME → dismissal, ending with `Successfully closed creative — PASS (providerCleared=YES, stillAttachedToParent=NO)`. No spurious teardown logs.
- [x] **Timeout / non-rendering case** — Switched domain to `76ers` (creative does not render for this account, so JS iframe-detection times out after 5s). Log shows exactly:
  ```
  Creative not opened: JS iframe-detection timed out
  Teardown after JS iframe-detection timed out — PASS (providerCleared=YES, stillAttachedToParent=NO)
  ```
  No ghost second `.notOpened`, no leftover WebView in the parent's subview tree, state back to `.closed`. Native 5s timer fired afterward but the epoch guard silently dropped it (as designed).
- [x] Verified interactivity — app is fully touch-responsive immediately after teardown; no invisible full-screen WebView eating events.

## Platform / OS coverage

- iOS 26.1 simulator (iPhone 17 Pro)
- Deployment target: iOS 14.0 (no new availability gates)

## Public API impact

None. `closeCreative` stays internal. The only observable behavior change is that `.notOpened` is now followed by teardown instead of a lingering WebView; the callback contract (single `.notOpened` per failed launch, no trailing `.closed`) is preserved and covered by a regression test.

## Host-app rollback

Pure revert. No persisted state, no schema changes, no new public symbols. Reverting this commit range restores prior behavior.


Sources/ATTNWebViewHandling.swift
```diff
@@ -236,33 +306,64 @@ extension ATTNWebViewHandler: WKNavigationDelegate {
...
             }
         }
     }
 
+    func webView(_ webView: WKWebView, didFail navigation: WKNavigation!, withError error: Error) {
+        let epoch = (webView as? CustomWebView)?.launchEpoch ?? 0
+        reportNotOpenedAndTearDown(epoch: epoch, reason: "navigation failed: \(error.localizedDescription)")
+    }
+
+    func webView(_ webView: WKWebView, didFailProvisionalNavigation navigation: WKNavigation!, withError error: Error) {
```
usharif-attentive: **[High] This will tear down legitimate launches on benign cancellations.** WebKit delivers `didFailProvisionalNavigation` with `NSURLErrorCancelled` (-999) when a provisional load is superseded — e.g. the creative page issues an immediate JS/meta redirect while the initial load is still provisional — and with `WKError` 102 ("frame load interrupted by policy change") when `decidePolicyFor` returns `.cancel`, which this SDK's own policy handler does for `sms:` and no-target-frame URLs. During `.launching`, epoch and state both match, so a redirect-during-load kills a creative that would have rendered.

Recommend filtering both codes in this handler and in `didFail` above before calling `reportNotOpenedAndTearDown`:

```swift
let nsError = error as NSError
if nsError.code == NSURLErrorCancelled || (nsError.domain == "WebKitErrorDomain" && nsError.code == 102) {
    return
}
```

Sources/ATTNWebViewHandling.swift
```diff
@@ -190,6 +204,60 @@ class ATTNWebViewHandler: NSObject, ATTNWebViewHandling {
     }
 
     func closeCreative() {
```
usharif-attentive, line 202: **[Medium] `closeCreative` can emit `.closed` twice via `onRemovedFromWindow`.** `detachAndNullWebView` removes the webview on main while the `.open → .closed` state update sits in a separate `creativeQueue.async`. If the removal lands first (likely when `closeCreative` is called from main, e.g. the CLOSE script message), `didMoveToWindow` fires synchronously, sees state still `.open`, and re-enters `closeCreative` — a second `.closed` reaches the host.

This race is pre-existing, but the failure path got a CAS guard in this PR while this path didn't — and the PR adds a new caller (`webViewWebContentProcessDidTerminate`). A `compareAndSet(from: .open, to: .closed)` guard here would close it the same way. Fine as a fast follow-up if you'd rather keep this PR scoped.

Sources/ATTNWebViewHandling.swift
```diff
@@ -190,6 +204,60 @@ class ATTNWebViewHandler: NSObject, ATTNWebViewHandling {
...
+                handler?(ATTNCreativeTriggerStatus.notOpened)
+            }
+            self.detachAndNullWebView(logContext: "Teardown after \(reason)")
+        }
+    }
+
+    /// JS reported the creative iframe is present. Gate the `.launching → .open`
+    /// transition on the epoch check so a stale JS-success from an old webview
+    /// can't flip a new launch's state and disable its native timeout guard.
+    private func handleJSSuccess(epoch: UInt64) {
```
usharif-attentive, line 243: **[Low] Note the `.open` semantics shift.** State now transitions `.launching → .open` at JS iframe-detection, whereas before `.open` was only set on the `IMPRESSION` message. That's the intended fix for the timer-kills-healthy-creative race, but it also changes behavior of state-gated message handling (e.g. the `document-visibility: true where state == .open` suppression case) in the window between JS-success and IMPRESSION. Probably fine — worth a sentence in the PR description.

Sources/ATTNWebViewHandling.swift
```diff
@@ -199,14 +267,16 @@ class ATTNWebViewHandler: NSObject, ATTNWebViewHandling {
...
-            guard let self = self else { return }
-            stateManager.updateState(.closed)
-            self.webViewProvider?.triggerHandler?(ATTNCreativeTriggerStatus.closed)
-            Loggers.creative.debug("Successfully closed creative - Visitor ID: \(self.userIdentity.visitorId, privacy: .public)")
+            // Verify the detach actually happened. If this ever logs FAIL, the
+            // WebView is leaking somewhere and we have a bug.
+            let providerCleared = (self.webViewProvider?.webView == nil)
+            let stillAttached = tornDownWebView.map { tornDownParent?.subviews.contains($0) ?? false } ?? false
+            let passed = providerCleared && !stillAttached
+            Loggers.creative.debug("\(logContext, privacy: .public) — \(passed ? "PASS" : "FAIL", privacy: .public) (providerCleared=\(providerCleared ? "YES" : "NO", privacy: .public), stillAttachedToParent=\(stillAttached ? "YES" : "NO", privacy: .public))")
```
usharif-attentive, line 285: Nit: the old "Successfully closed creative" log carried the Visitor ID; this replacement dropped it. The rest of the file logs it consistently — worth restoring for support escalations (which is exactly what spawned this PR).

Tests/TestCases/ATTNWebViewHandlerTests.swift
```diff
@@ -237,9 +395,58 @@ class MockCreativeUrlProvider: ATTNCreativeUrlProviding {
...
         onMakeWebView?(webView)
         return webView
     }
 }
+
+/// Test handler that stubs the WKNavigationDelegate.didFinish path by short-circuiting
+/// the JS evaluation call. When `stubbedJSResult` is set, the didFinish call skips
+/// callAsyncJavaScript and dispatches straight into the SDK's result-handling switch,
+/// mirroring the production behavior (fire .notOpened, then tear down the webview).
+class JSTimeoutStubHandler: ATTNWebViewHandler {
```
usharif-attentive, line 410: **This stub means the JS-timeout test doesn't exercise production code.** `JSTimeoutStubHandler` overrides `didFinish` and reimplements teardown in `reportNotOpenedAndTearDownForTest`, so `testDidFinish_JSTimeout_...` asserts the behavior of test-only code — the real `callAsyncJavaScript` completion branch (epoch read + `reportNotOpenedAndTearDown`) never runs.

Since `webViewProvider` was already widened to `internal` for tests, consider making `reportNotOpenedAndTearDown` internal too so the stub can call the real primitive, keeping only the JS evaluation stubbed.

Tests/TestCases/ATTNWebViewHandlerTests.swift
```diff
@@ -121,6 +121,164 @@ final class ATTNWebViewHandlerIntegrationTests: XCTestCase {
...
+        DispatchQueue.main.asyncAfter(deadline: .now() + 7.0) {
+            allDone.fulfill()
+        }
+        wait(for: [allDone], timeout: 10.0)
+
+        XCTAssertEqual(notOpenedCount, 1, ".notOpened must fire exactly once even if both timeout paths run")
+        XCTAssertTrue(otherStatuses.isEmpty, "must not additionally emit .closed or .opened on the timeout path; got \(otherStatuses)")
+    }
+
+    func testMultiLaunch_StaleDelegateCallback_DoesNotTearDownNewLaunch() {
```
usharif-attentive: **This test manufactures its own epoch mismatch rather than testing the real mechanism.** It abandons the same-handler scenario partway (the comments narrate the aborted approach), builds a second handler, then force-sets `webView1.launchEpoch = 0` to create the mismatch. So it verifies "0 ≠ 1 is dropped," not that two launches on one handler naturally produce distinct epochs — which is the actual bug being guarded.

Reusing the same handler and provider for both launches (resetting the mock's expectations between them) would exercise the real cross-launch flow: webView1 carries epoch 1, `currentLaunchEpoch` is 2, and the stale callback is dropped without any forced values. The dead-end comments and `_ = epoch1 // silence unused warning` should go regardless.

Also worth noting: the new `didFail`, `webViewWebContentProcessDidTerminate` (both states), and `onRemovedFromWindow`-during-`.launching` paths have no positive-case tests — all are directly callable delegate methods.

Tests/TestCases/ATTNWebViewHandlerTests.swift
```diff
@@ -121,6 +121,164 @@ final class ATTNWebViewHandlerIntegrationTests: XCTestCase {
         }
     }
 
+    func testLaunchCreative_NativeTimeout_TearsDownWebViewAndReportsNotOpened() {
```
usharif-attentive, line 125: Nit: the three new tests lean on the real 5s timer plus fixed sleeps (~20s added to the suite). Injecting `timeoutInterval` as a constructor parameter with the 5.0 default would make these near-instant.

docs/superpowers/plans/2026-08-03-msdk-450-webview-timeout-teardown.md
```diff
@@ -0,0 +1,540 @@
+# MSDK-450: Guaranteed WebView Teardown on Timeout — Implementation Plan
```
usharif-attentive: **This planning doc shouldn't ship in the public repo.** CLAUDE.md explicitly keeps internal process details out of the public SDK repo, and this contains internal agent-workflow instructions. It's also stale relative to the code it accompanies: it describes a `didTimeOut: Bool` design that the implementation replaced with the epoch scheme, so it would actively mislead a future reader. Recommend dropping it from the PR.

usharif-attentive (commented):
Overall the fix is well-designed — the epoch + CAS combination is the right shape for this class of stale-callback race, and routing all failure paths through one teardown primitive is a solid structure. Three things I'd want addressed before merge (details inline): (1) the new `didFail*` handlers will tear down legitimate launches on benign cancellations (NSURLErrorCancelled / WKError 102) — that one can regress working creatives; (2) the internal planning doc shouldn't ship in this public repo; (3) two of the new tests validate test-double code rather than the production paths. The `closeCreative` double-`.closed` race is pre-existing and fine as a fast follow-up.

Sources/ATTNWebViewHandling.swift
adelag-attentive: Thanks for flagging. After investigation, we're going the **stricter** route: removing `didFail` and `didFailProvisionalNavigation` overrides entirely rather than filtering codes inside them.

Reasoning:
- Confirmed with `frontend-code` (`libs/tag-mobile-apps/src/index.ts`, `libs/tag-creatives`) that the mobile-apps creative page does **no** top-level redirects — no `location.href`, `location.replace`, `meta http-equiv=refresh`, or main-frame supersessions. The `multiPage/index.html` load is iframe-only.
- The concrete source of these benign codes in production is our own `decidePolicyFor` cancellations (`sms:` and no-target-frame URLs). Never hooking the failure delegates in the first place removes that entire trap without needing a filter that could go stale.
- The native 5s launching timeout backstops real hard failures (DNS, TLS, 4xx/5xx) — the iframe never appears → timeout fires → teardown. Slower by up to 5s in those edge cases, but no risk of tearing down a healthy creative.

Fixed in 393a38d.

Sources/ATTNWebViewHandling.swift
adelag-attentive: Good catch. Added a guard in `closeCreative`: if state is already `.closed`, no-op. Prevents re-entry via `onRemovedFromWindow` from firing `.closed` twice or racing on teardown. Fixed in 393a38d.

Sources/ATTNWebViewHandling.swift
adelag-attentive: You're right — the transition is still there. Kept because it closes the last remaining premature-teardown scenario after the scope narrowing: without it, the native 5s timer can still fire between JS-success (`.opened`) and IMPRESSION arriving, and tear down a rendered creative.

The visibility-event semantics shift you noted is real: the `document-visibility: true where state == .open` suppression branch will now activate a few ms earlier (JS-success instead of IMPRESSION). Read that code path carefully; the branch is a no-op except for a log line (`ATTNWebViewHandling.swift:441`), so no functional change. Flagged in the PR description now for reviewers.

Sources/ATTNWebViewHandling.swift
adelag-attentive: Restored in 393a38d — the `Successfully closed creative` log line now includes the Visitor ID again.

Tests/TestCases/ATTNWebViewHandlerTests.swift
adelag-attentive: Great point. Fixed in 393a38d — made `reportNotOpenedAndTearDown` internal, and `JSTimeoutStubHandler` now calls the real method with the actual webview's epoch. The test also now explicitly invokes `didFinish` on the handler (since `MockWKWebView.load` returns nil and no real navigation fires), so it exercises the full `didFinish → reportNotOpenedAndTearDown → detachAndNullWebView` pipeline.

Tests/TestCases/ATTNWebViewHandlerTests.swift
adelag-attentive: Rewrote in 393a38d as `testMultiLaunch_StaleTimeoutCallback_DoesNotTearDownNewLaunch`. It now uses ONE handler naturally: launch #1 → close → launch #2, then invokes `handler.reportNotOpenedAndTearDown(epoch: 1, ...)` to simulate launch #1's stale native timer firing after launch #2 has advanced the epoch. No forced values, no aborted-approach comments.

On the missing positive-case tests for the removed delegate methods: those overrides are gone entirely as of 393a38d, so no test coverage is needed for them. `onRemovedFromWindow`-during-`.launching` is likewise back to pre-PR behavior (unconditional `closeCreative`).

Tests/TestCases/ATTNWebViewHandlerTests.swift
adelag-attentive: Fixed in 393a38d — `launchTimeoutInterval` is now a constructor parameter (default 5.0). Tests run with 0.2s. The WebViewHandlerIntegrationTests suite dropped from ~20s to ~1.3s.

docs/superpowers/plans/2026-08-03-msdk-450-webview-timeout-teardown.md
adelag-attentive: Agreed. Deleted the doc in 393a38d — it was stale (described the pre-refactor `didTimeOut: Bool` design that got replaced with the epoch scheme) and doesn't belong in the public repo per CLAUDE.md.

Sources/ATTNWebViewHandling.swift
```diff
@@ -88,6 +104,8 @@ class ATTNWebViewHandler: NSObject, ATTNWebViewHandling {
 
         creativeQueue.async { [weak self] in
             guard let self = self else { return }
+            self.currentLaunchEpoch &+= 1
```
usharif-attentive, line 107: **Non-blocking — epoch increment isn't atomic with the state CAS.** The `.closed → .launching` CAS runs synchronously on the caller's thread (line 100), but this increment runs later on `creativeQueue`. If a host retriggers a creative in the instant a stale timer block from the previous launch is executing on the queue — after the new launch's CAS but before this increment — the stale timer sees a matching epoch *and* `.launching` state, and spuriously fires `.notOpened` + teardown against the new launch. Its webview then gets created under `.closed` state and lingers (until the next launch's prevent-dupes block removes it).

The window is microseconds wide and requires the retrigger to land exactly at the previous timeout's deadline, so this is follow-up material, not a blocker. The tightening is small: move the CAS inside the `creativeQueue.async` block so it sits next to the increment, making state transition + epoch bump atomic with respect to the queue.

Sources/ATTNWebViewHandling.swift
```diff
@@ -190,6 +200,70 @@ class ATTNWebViewHandler: NSObject, ATTNWebViewHandling {
...
+    }
+
+    /// Reports `.notOpened` and tears down the WebView, but only if `epoch` matches
+    /// the current launch. A stale callback from a previous launch (e.g. a not-yet-
+    /// fired 5s asyncAfter, a late JS completion after stopLoading) is silently
+    /// dropped so it can't tear down a healthy WebView belonging to a subsequent
+    /// launch.
+    /// Internal (rather than private) so test doubles can drive the real teardown
+    /// path from stubbed delegate methods.
+    func reportNotOpenedAndTearDown(epoch: UInt64, reason: String) {
```
usharif-attentive, line 226: **Non-blocking — crash-while-open is uncovered after the scope narrowing.** With `webViewWebContentProcessDidTerminate` removed, a WebContent process death while a creative is displayed leaves a dead, blank webview attached with no teardown path (this CAS only fires from `.launching`, and the 5s timer is long gone by then). That matches `main`'s behavior, so it's not a regression — but the customer report explicitly named WebContent crashes, so it deserves a follow-up ticket rather than silence. Restoring the delegate method with the `.open → closeCreative` / `.launching → reportNotOpenedAndTearDown` split from the previous revision (d930487) would cover it whenever that ticket gets picked up.

usharif-attentive (commented):
Re-verified the branch at 393a38d: ran the full suite locally (227 tests, 0 failures, iPhone 17 Pro / iOS 26.1), lint clean, CI green, and confirmed all previous review threads are addressed in code. Two non-blocking observations inline — neither should hold the merge.

usharif-attentive (approved):

PR #292, opened 2026-08-04
MSDK-312: move visitor-id logging outside the identity lock

usharif-attentive:
## Summary

Fixes the recurring CI timeout in `testClearUser_concurrentClearAndMerge_doesNotCrash` (MSDK-312) at its root: `clearUser()` was holding the identity `NSLock` across an `os_log` call, serializing all concurrent identity operations at log-capture speed.

## Key Changes

`ATTNUserIdentity.clearUser()` calls `ATTNVisitorService.createNewVisitorId()` inside its lock, and that method logged the new visitor id via `os_log`. Logging is I/O with unbounded latency — CircleCI's log capture serializes it at roughly 75ms per call — so every racing `mergeIdentifiers()` on other threads stalled behind the log while the lock was held. This is why the concurrency test kept timing out on CI despite three prior mitigations (mocked storage, raised timeout, halved iterations), and it's a genuine contention hazard for any host app calling identity APIs from multiple threads.

The fix moves logging out of the critical section: `createNewVisitorId()` now only generates and persists (documented as deliberately log-free), and the "Generated new visitor id" line is emitted by the call sites — `getVisitorId()`'s new-id path, and `clearUser()` after the lock is released. The `UserDefaults` write stays inside the lock, preserving the existing invariant that disk write order matches the in-memory swap order.

With the bottleneck gone, the concurrency test is restored to 200 iterations (reverting the earlier shrink to 50).

## Verification

- Full unit suite locally: 223 tests, 0 failures (iPhone 17 Pro simulator, iOS 26.1, Xcode 26.6).
- `testClearUser_concurrentClearAndMerge_doesNotCrash` now completes in **0.010s at 200 iterations**; before this change it timed out at 15s with only 50 iterations on CI (pipeline 591, job 4594), confirming the lock-held log was the entire cost.
- CI on this PR re-runs the suite including the restored 200-iteration test.

## Risk

Low. No public API changes — method signatures, behavior, log messages, and log category are unchanged; only where the log statement executes relative to the lock moved. The one observable difference is log ordering: under concurrent `clearUser()` calls, "Generated new visitor id" lines may interleave in a different order than the actual id assignment. The persisted-id invariant (disk order matches memory order) is untouched. Binary and source compatible for host apps.

## Observability

No direct SDK telemetry on this path. If this regressed, we'd see it as: threading/lock crashes in ATTN symbols via host-app crash reporters (Crashlytics/Sentry), or visitor-id anomalies in backend event ingestion / identity resolution dashboards.

## Rollback

No feature flag on this path — rollback is revert + patch release, gated on host-app adoption. Acceptable here because the change is a pure internal refactor with no payload, persistence-format, or API changes.

### DIFF
--- a/Sources/Public/ATTNUserIdentity.swift
+++ b/Sources/Public/ATTNUserIdentity.swift
@@ -54,13 +54,13 @@ public final class ATTNUserIdentity: NSObject {
         // a stale visitor id. Logging stays outside: os_log latency is
         // unbounded, and holding the lock across it stalls every other
         // identity call on other threads.
-        let newVisitorId: String = lock.withLock {
+        let newVisitorId = lock.withLock { () -> String in
             _identifiers = [:]
-            let newVisitorId = visitorService.createNewVisitorId()
-            _visitorId = newVisitorId
-            return newVisitorId
+            let id = visitorService.createNewVisitorId()
+            _visitorId = id
+            return id
         }
-        Loggers.event.info("Generated new visitor id: \(newVisitorId, privacy: .public)")
+        visitorService.logNewVisitorId(newVisitorId)
     }
 
     @objc
--- a/Sources/Public/ATTNUserIdentity.swift
+++ b/Sources/Public/ATTNUserIdentity.swift
@@ -51,11 +51,16 @@ public final class ATTNUserIdentity: NSObject {
         // order matches the in-memory swap order. If two threads race here and
         // we generate outside the lock, the last write to disk could disagree
         // with the last value of `_visitorId`, leaving the next app launch with
-        // a stale visitor id.
-        lock.withLock {
+        // a stale visitor id. Logging stays outside: os_log latency is
+        // unbounded, and holding the lock across it stalls every other
+        // identity call on other threads.
+        let newVisitorId: String = lock.withLock {
             _identifiers = [:]
-            _visitorId = visitorService.createNewVisitorId()
+            let newVisitorId = visitorService.createNewVisitorId()
+            _visitorId = newVisitorId
+            return newVisitorId
         }
+        Loggers.event.info("Generated new visitor id: \(newVisitorId, privacy: .public)")
     }
 
     @objc
--- a/Tests/TestCases/ATTNUserIdentityTests.swift
+++ b/Tests/TestCases/ATTNUserIdentityTests.swift
@@ -6,6 +6,7 @@
 //
 
 import XCTest
+import os
 @testable import ATTNSDKFramework
 
 final class ATTNUserIdentityTests: XCTestCase {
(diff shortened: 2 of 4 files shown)


usharif-attentive:
@codex review

usharif-attentive:
@attentive-claude review

Sources/Public/ATTNUserIdentity.swift
```diff
@@ -51,11 +51,16 @@ public final class ATTNUserIdentity: NSObject {
...
+        // a stale visitor id. Logging stays outside: os_log latency is
+        // unbounded, and holding the lock across it stalls every other
+        // identity call on other threads.
+        let newVisitorId: String = lock.withLock {
             _identifiers = [:]
-            _visitorId = visitorService.createNewVisitorId()
+            let newVisitorId = visitorService.createNewVisitorId()
+            _visitorId = newVisitorId
+            return newVisitorId
         }
```
adelag-attentive, line 62: Nit: the outer `let newVisitorId: String = lock.withLock` and this inner `let newVisitorId = visitorService.createNewVisitorId()` share a name. It compiles cleanly since they live in different scopes, but reads a little noisy. Consider renaming the inner one to `id` (or leaning on the closure’s implicit return):

```swift
let newVisitorId = lock.withLock { () -> String in
    _identifiers = [:]
    let id = visitorService.createNewVisitorId()
    _visitorId = id
    return id
}
```

Sources/Public/ATTNUserIdentity.swift
```diff
@@ -51,11 +51,16 @@ public final class ATTNUserIdentity: NSObject {
...
+        Loggers.event.info("Generated new visitor id: \(newVisitorId, privacy: .public)")
     }
```
adelag-attentive, line 65: Nit (DRY / fragile contract): the `"Generated new visitor id: ..."` log line is now duplicated verbatim here and in `ATTNVisitorService.getVisitorId()`’s new-id branch, and `createNewVisitorId()` is silently non-logging — a future caller who forgets the doc comment will drop the log entirely.

Consider extracting a small internal helper so the message stays in one place and the "must log after create" contract is centralized, e.g.:

```swift
// in ATTNVisitorService
func createAndLogNewVisitorId() -> String {
    let id = createNewVisitorId()
    Loggers.event.info("Generated new visitor id: \(id, privacy: .public)")
    return id
}
```

Then `getVisitorId()` calls the logging variant, and `clearUser()` calls `createNewVisitorId()` inside the lock and `Loggers.event.info(...)` (or a shared log helper) outside. Not blocking — just reduces drift risk.

Tests/TestCases/ATTNUserIdentityTests.swift
```diff
@@ -111,17 +111,16 @@ final class ATTNUserIdentityTests: XCTestCase {
...
+        // doesn't dominate wall clock — the test asserts the lock, not disk I/O.
+        // clearUser() logs outside its critical section, so iterations can be
+        // high enough to exercise real contention without CI log-capture
+        // latency serializing the whole test.
         let identity = ATTNUserIdentity(
             identifiers: [:],
             visitorService: ATTNVisitorService(persistentStorage: ATTNPersistentStorageMock())
         )
-        runConcurrently(iterations: 50, timeout: 15, queueLabels: ["merge", "clear"]) { i, queueIndex in
+        runConcurrently(iterations: 200, timeout: 15, queueLabels: ["merge", "clear"]) { i, queueIndex in
```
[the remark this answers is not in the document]
usharif-attentive: Good catch — the log call was indeed still on the timed path (`group.leave()` only fires after the block returns), so CI log-capture serialization could have re-flaked this at 200 iterations. Addressed in 3e8f1e6: `ATTNVisitorService` now takes an injectable `Logger` (defaulting to `Loggers.event`) and this test injects `Logger(OSLog.disabled)`, so no os_log happens on the timed path at all.

Sources/Public/ATTNUserIdentity.swift
usharif-attentive: Done in 3e8f1e6 — inner binding renamed to `id` with the closure typed `() -> String`, matching your suggestion.

Sources/Public/ATTNUserIdentity.swift
usharif-attentive: Done in 3e8f1e6 — the message now lives in one place: `ATTNVisitorService.logNewVisitorId(_:)`. `getVisitorId()`'s new-id branch and `clearUser()` both call it, and `createNewVisitorId()`'s doc comment points to it as the explicit logging counterpart. I kept create and log as separate calls (rather than a combined `createAndLogNewVisitorId()`) because `clearUser()` must run the create inside the lock and the log outside it.

adelag-attentive (dismissed):

adelag-attentive (approved):

PR #294, opened 2026-08-07
MSDK-464: expose inbox unread count synchronously for UIKit; Added Inbox Readme

adelag-attentive:
Jira: [MSDK-464](https://attentivemobile.atlassian.net/browse/MSDK-464)

## What

Existing inbox unread APIs (`unreadCount`, `inboxStateStream`) require async/await and materialization on the `InboxManager` actor, which forces every Swift UIKit host that just wants an unread badge to spawn a `Task` loop and `await` on the actor. That's the one place the drop-in inbox story leaks Swift Concurrency out to callers who otherwise never need to touch it.

This PR adds an **additive, synchronous mirror** alongside the existing async surface — no signature or behavior change to any existing API. **Swift only** — consistent with the rest of the new inbox API surface, which is also Swift-only.

## New public surface

| Symbol | Kind |
|---|---|
| `ATTNSDK.inboxUnreadCount: Int` | Synchronous, thread-safe (Swift) |
| `Notification.Name.ATTNSDKInboxUnreadCountChanged` | Posted on every count change |

- **`inboxUnreadCount`** — synchronous read of the server-authoritative count. Returns `0` until the first fetch completes, matching the async `unreadCount` accessor's initial-value semantics.
- **`.ATTNSDKInboxUnreadCountChanged`** — `object` is the `ATTNSDK` instance (filter by this for multi-instance hosts; pass `nil` to observe all); `userInfo["unreadCount"]` is the new `Int`. Same-value writes are **deduped** so a re-fetch that returns the same count does not fire a notification and the badge never churns.

## Swift UIKit usage (no async)

```swift
override func viewDidLoad() {
    super.viewDidLoad()

    // Paint initial state synchronously — no Task, no await
    updateBadge(sdk.inboxUnreadCount)

    NotificationCenter.default.addObserver(
        forName: .ATTNSDKInboxUnreadCountChanged,
        object: sdk,
        queue: .main
    ) { [weak self] note in
        let count = note.userInfo?["unreadCount"] as? Int ?? 0
        self?.updateBadge(count)
    }

    // Kick off a fetch so the notification path has something to deliver.
    // The README recommends calling refreshInboxUnreadCount() on app launch
    // and after a push open regardless.
    Task { await sdk.refreshInboxUnreadCount() }
}
```

## Implementation

- New `UnreadCountBox` (nonisolated, lock-guarded `Int` + `onChange` callback) in `Sources/Inbox/InboxManager.swift`. Reads use the SDK's existing `NSLock.withLock` helper. Same-value dedupe lives inside the box. `onChange` fires **after** the lock is released so a re-entrant read from inside the callback does not deadlock.
- `InboxManager.storedUnreadCount` gets a one-line `didSet` that mirrors every write into the box — catches all 9 existing write sites (fetch, mark-read/unread PATCH response, delete, optimistic ±1s, identity reset) with zero change to their code paths.
- `ATTNSDK` owns the box; its `onChange` posts `.ATTNSDKInboxUnreadCountChanged`. Eagerly initialized in `init` (not `lazy var`) because `lazy` init is not synchronized and `inboxUnreadCount` is documented thread-safe.
- `materializedInboxManager()` passes the SDK's box into the manager so all writes route back to the mirror.

## Public API impact

Additive only. Zero changes to:
- Method or property signatures
- Behavior of any existing API (`unreadCount`, `inboxStateStream`, `refreshInboxUnreadCount`, `markRead/markUnread/delete`, `inboxView`, etc.)
- Threading contracts, error types, or persistence format

Host apps that never touch the new APIs are byte-for-byte unaffected at runtime.

## Verification

- **Unit tests: 366/366 passing** (5 new, in `Tests/TestCases/InboxManagerTests.swift`):
  - Initial fetch publishes the count to the sync mirror.
  - Same-value re-fetch does **not** re-fire `onChange`.
  - `markRead` decrements the mirror correctly (optimistic + server reconcile).
  - `resetForIdentityChange` zeroes the mirror.
  - Re-entrant `count` read from inside `onChange` does **not** deadlock.
- **Framework build: clean** (`xcodebuild -project attentive-ios-sdk.xcodeproj -scheme attentive-ios-sdk-framework -sdk iphonesimulator`).
- **Bonni build: clean.**
- **Manual verification on simulator (iPhone 17 Pro):** Temporarily swapped Bonni's existing red-badge UI to read from `sdk.inboxUnreadCount` and observe `.ATTNSDKInboxUnreadCountChanged` under the hood (same UI, only the plumbing changed). Confirmed the unread count updates correctly across:
  - Initial fetch on app launch
  - `triggerCreative` → mark-read on tap
  - Swipe-right to mark-read / mark-unread on the inbox row
  - Swipe-left to delete
  - Identity reset (`clearUser`)
  - No lag or drift vs. the AsyncStream path

  Bonni is reverted to the AsyncStream integration in this PR — no host-side changes ship. The manual test just proved the new plumbing works end-to-end on a real host app.

## README

New section under **Inbox → Show an unread badge** documents both patterns:
- **UIKit (no async/await)** — the new sync + NotificationCenter path.
- **SwiftUI / async-first** — the existing `inboxStateStream` pattern, unchanged.

[MSDK-464]: https://attentivemobile.atlassian.net/browse/MSDK-464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Helpers/Extension/Notification+Extension.swift
+++ b/Sources/Helpers/Extension/Notification+Extension.swift
@@ -25,7 +25,7 @@ extension Notification.Name {
         ///
         /// - `object`: the `ATTNSDK` instance that owns the count. Filter by this when running
         ///   multiple SDK instances; pass `nil` to observe all instances.
-        /// - `userInfo["unreadCount"]`: `Int` — the new server-authoritative count.
+        /// - `userInfo["attentiveInboxUnreadCount"]`: `Int` — the new server-authoritative count.
         ///
         /// The notification is dispatched from whichever thread wrote the count. Observers that
         /// touch UIKit should register with `queue: .main` or hop to main before reading.
--- a/Sources/Helpers/Extension/Notification+Extension.swift
+++ b/Sources/Helpers/Extension/Notification+Extension.swift
@@ -19,9 +19,9 @@ extension Notification.Name {
         ///   - `"attentiveInboxActionUrl"`: `URL` — present only when the message has a valid `actionURL`
         public static let ATTNSDKInboxMessageTapped = Notification.Name("ATTNSDKInboxMessageTapped")
 
-        /// Posted whenever the inbox unread count changes. Provides a UIKit / Objective-C-friendly
-        /// alternative to `ATTNSDK.inboxStateStream` for hosts driving an unread badge without
-        /// adopting Swift Concurrency.
+        /// Posted whenever the inbox unread count changes. Provides a UIKit-friendly alternative
+        /// to `ATTNSDK.inboxStateStream` for Swift hosts driving an unread badge without adopting
+        /// Swift Concurrency.
         ///
         /// - `object`: the `ATTNSDK` instance that owns the count. Filter by this when running
         ///   multiple SDK instances; pass `nil` to observe all instances.
--- a/Sources/Helpers/Extension/Notification+Extension.swift
+++ b/Sources/Helpers/Extension/Notification+Extension.swift
@@ -18,4 +18,17 @@ extension Notification.Name {
         ///   - `"attentiveInboxMessageId"`: `Message.ID` (String) — always present
         ///   - `"attentiveInboxActionUrl"`: `URL` — present only when the message has a valid `actionURL`
         public static let ATTNSDKInboxMessageTapped = Notification.Name("ATTNSDKInboxMessageTapped")
+
+        /// Posted whenever the inbox unread count changes. Provides a UIKit / Objective-C-friendly
+        /// alternative to `ATTNSDK.inboxStateStream` for hosts driving an unread badge without
+        /// adopting Swift Concurrency.
+        ///
+        /// - `object`: the `ATTNSDK` instance that owns the count. Filter by this when running
+        ///   multiple SDK instances; pass `nil` to observe all instances.
+        /// - `userInfo["unreadCount"]`: `Int` — the new server-authoritative count.
+        ///
+        /// The notification is dispatched from whichever thread wrote the count. Observers that
+        /// touch UIKit should register with `queue: .main` or hop to main before reading.
+        /// Same-value writes are deduped, so a re-fetch that returns the same count does not fire.
+        public static let ATTNSDKInboxUnreadCountChanged = Notification.Name("ATTNSDKInboxUnreadCountChanged")
 }
--- a/Sources/Inbox/InboxManager.swift
+++ b/Sources/Inbox/InboxManager.swift
@@ -666,6 +666,12 @@ extension InboxManager {
 /// or a NotificationCenter-driven badge — powers `ATTNSDK.inboxUnreadCount` and the
 /// `ATTNSDKInboxUnreadCountChanged` fanout. Same-value writes are deduped. `onChange` runs after
 /// the lock is released so a re-entrant read from inside the callback does not deadlock.
+///
+/// **Ordering invariant:** `onChange` callbacks are only guaranteed to arrive in write-order when
+/// there is a single serialized writer. In this SDK that writer is the `InboxManager` actor's
+/// `didSet` on `storedUnreadCount` (see the property observer above), which is safe because
+/// actor isolation serializes all writes. A second concurrent writer would race post-unlock and
+/// could deliver stale counts out of order — do not introduce one without gating the callback.
 final class UnreadCountBox: @unchecked Sendable {
     private let lock = NSLock()
     private var storedCount: Int = 0
--- a/Sources/Inbox/InboxManager.swift
+++ b/Sources/Inbox/InboxManager.swift
@@ -60,7 +60,14 @@ actor InboxManager {
     /// would undercount unread messages that live on unfetched pages. Reads should go through the
     /// `unreadCount` accessor, which awaits the init-time refresh so the first read never returns
     /// a stale 0.
-    private var storedUnreadCount: Int = 0
+    ///
+    /// `didSet` mirrors every write into `unreadCountBox`; see `UnreadCountBox` for semantics.
+    private var storedUnreadCount: Int = 0 {
+        didSet { unreadCountBox.update(to: storedUnreadCount) }
+    }
+
+    /// Nonisolated mirror of `storedUnreadCount`; see `UnreadCountBox` below.
+    private let unreadCountBox: UnreadCountBox
 
     private let api: ATTNAPIProtocol
     private let identityProvider: InboxIdentityProvider
@@ -166,9 +173,14 @@ actor InboxManager {
         }
     }
 
-    init(api: ATTNAPIProtocol, identityProvider: @escaping InboxIdentityProvider) {
+    init(
+        api: ATTNAPIProtocol,
+        identityProvider: @escaping InboxIdentityProvider,
+        unreadCountBox: UnreadCountBox = UnreadCountBox()
+    ) {
         self.api = api
         self.identityProvider = identityProvider
+        self.unreadCountBox = unreadCountBox
         // Fetch the first page of messages and the unread count on construction so passive-badge
         // hosts (`sdk.unreadCount`) and stream/`allMessages` consumers both resolve without
         // needing to present `inboxView()`. Non-inbox hosts never construct the manager — see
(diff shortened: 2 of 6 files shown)


Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -59,6 +59,18 @@ public final class ATTNSDK: NSObject {
...
+    /// in `init` — `lazy var`'s initializer is not synchronized, so two concurrent first-reads of
+    /// `inboxUnreadCount` could construct two boxes and later reads (or the manager's stored
+    /// reference) would see the wrong one. `private var` (not `let`) because Swift disallows
+    /// referencing `self` from a stored-property initializer, and the `[weak self]` capture in
+    /// `onChange` needs it — so the value is set once, after `super.init()`, and never rewritten.
+    /// The weak capture prevents a retain cycle (SDK → box → onChange → SDK).
+    private var unreadCountBox: UnreadCountBox!
+
+    /// `@objc` synchronous mirror of `unreadCount`; see `.ATTNSDKInboxUnreadCountChanged`.
+    @objc public var inboxUnreadCount: Int { unreadCountBox.count }
```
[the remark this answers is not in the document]
adelag-attentive: Thanks — but we don't support Objective-C for this API. The rest of the SDK has ObjC-visible surfaces for backwards compatibility, but new inbox APIs (`inboxStateStream`, `unreadCount`, `refreshInboxUnreadCount`, `inboxView`, `markRead/markUnread/delete`) are all Swift-only — this one is intentionally aligned with that.

For a Swift host, the fetch is triggered by any of: `await sdk.refreshInboxUnreadCount()` (the pattern the README recommends on app launch and after a push open), materializing `sdk.inboxView()` / `inboxViewController()`, or subscribing to `sdk.inboxStateStream`. `inboxUnreadCount` is a passive mirror by design, so hosts using this can start with any of the existing materialization paths and then observe the notification.

Will drop the "Objective-C" framing from the README section and the doc comment on the notification name in a follow-up commit so we don't imply an unsupported audience.

Sources/Helpers/Extension/Notification+Extension.swift
```diff
@@ -18,4 +18,17 @@ extension Notification.Name {
...
+        /// adopting Swift Concurrency.
+        ///
+        /// - `object`: the `ATTNSDK` instance that owns the count. Filter by this when running
+        ///   multiple SDK instances; pass `nil` to observe all instances.
+        /// - `userInfo["unreadCount"]`: `Int` — the new server-authoritative count.
+        ///
+        /// The notification is dispatched from whichever thread wrote the count. Observers that
+        /// touch UIKit should register with `queue: .main` or hop to main before reading.
+        /// Same-value writes are deduped, so a re-fetch that returns the same count does not fire.
+        public static let ATTNSDKInboxUnreadCountChanged = Notification.Name("ATTNSDKInboxUnreadCountChanged")
```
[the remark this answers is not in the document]
adelag-attentive: Thanks — but we don't support Objective-C for this API. The whole new inbox surface is Swift-only, and this notification is deliberately consistent with that. It also matches how every other notification in this SDK is exported (`ATTNSDKDeepLinkReceived`, `ATTNSDKInboxMessageTapped` are also plain `static let` on `Notification.Name`, not top-level `NSNotificationName` constants).

Will remove the "Objective-C-friendly" phrasing from the doc comment so it doesn't imply an audience we're not committing to.

Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -134,6 +146,15 @@ public final class ATTNSDK: NSObject {
 
         super.init()
 
+        self.unreadCountBox = UnreadCountBox { [weak self] newCount in
+            guard let self = self else { return }
+            NotificationCenter.default.post(
+                name: .ATTNSDKInboxUnreadCountChanged,
+                object: self,
+                userInfo: ["unreadCount": newCount]
```
usharif-attentive: **Blocking:** the `"unreadCount"` userInfo key breaks the `attentiveInbox*` convention set by `ATTNSDKInboxMessageTapped` (`"attentiveInboxMessageId"`, `"attentiveInboxActionUrl"`) — and userInfo keys are frozen public contract the moment this ships. Suggest renaming to `"attentiveInboxUnreadCount"` and hoisting the string to a constant (per repo convention) before merge. Two-minute fix now, breaking change later.

Sources/Public/SDK/ATTNSDK.swift
```diff
@@ -920,9 +941,16 @@ fileprivate extension ATTNSDK {
...
-        })
+        // Passing `unreadCountBox` here (rather than letting the manager build its own default)
+        // is what routes every write to `storedUnreadCount` back to `inboxUnreadCount` and the
+        // `.ATTNSDKInboxUnreadCountChanged` notification.
+        let manager = InboxManager(
+            api: api,
+            identityProvider: { [weak self] in
+                self?.identityStore.snapshot() ?? InboxIdentitySnapshot(visitorId: "", pushToken: "", email: nil, phone: nil)
+            },
+            unreadCountBox: unreadCountBox
```
usharif-attentive, line 952: This line is what the whole feature hinges on, but the SDK-level wiring has no unit coverage — the new tests exercise `UnreadCountBox` through `InboxManager` directly, so `sdk.inboxUnreadCount`, the notification posting (correct `object`/`userInfo`), and the box being passed here are only covered by the manual Bonni run. One test with `expectation(forNotification: .ATTNSDKInboxUnreadCountChanged, object: sdk)` would lock down the glue as a regression guard.

Sources/Inbox/InboxManager.swift
```diff
@@ -647,3 +659,34 @@ extension InboxManager {
...
     }
 }
+
+// MARK: - Unread count box
+
+/// Nonisolated mirror of `InboxManager.storedUnreadCount` for hosts that need a synchronous read
+/// or a NotificationCenter-driven badge — powers `ATTNSDK.inboxUnreadCount` and the
+/// `ATTNSDKInboxUnreadCountChanged` fanout. Same-value writes are deduped. `onChange` runs after
+/// the lock is released so a re-entrant read from inside the callback does not deadlock.
+final class UnreadCountBox: @unchecked Sendable {
```
usharif-attentive, line 675: Non-blocking: since `onChange` fires outside the lock, notification *ordering* is only guaranteed under a single serialized writer. That holds today (the actor's `didSet` is the only writer), but nothing documents it — a future second write site could deliver stale counts out of order. Suggest one doc line stating the invariant. (Deliberately not asking for a mechanical fix: holding the lock reintroduces the re-entrancy deadlock you test against, and a generation check has a check-then-act hole.)

usharif-attentive (commented):
Three inline comments: one blocking (userInfo key naming — frozen contract after merge), one strong-should (SDK-level test coverage), one non-blocking doc suggestion.

Sources/Public/SDK/ATTNSDK.swift
adelag-attentive: Good catch — done in 24c40d7. Renamed to `"attentiveInboxUnreadCount"` in the post site, the doc comment on `ATTNSDKInboxUnreadCountChanged`, and the README snippet. Kept it inline rather than hoisting to a constant to match the existing convention (`"attentivePushDeeplinkUrl"`, `"attentiveInboxMessageId"`, `"attentiveInboxActionUrl"` are all inline strings today).

Sources/Public/SDK/ATTNSDK.swift
adelag-attentive: Agreed — done in 24c40d7. Added two SDK-level tests to `ATTNSDKTests`:

- `testInboxUnreadCount_notificationFiresWithSDKObjectAndPayload` — drives a real `refreshInboxUnreadCount()` and asserts (a) the notification fires with the SDK as `object`, (b) `userInfo["attentiveInboxUnreadCount"]` carries the fetched value, and (c) `sdk.inboxUnreadCount` reflects it synchronously afterwards. This is the regression guard on the box-passing wiring in `materializedInboxManager()`.
- `testInboxUnreadCount_objectFilterIsolatesSDKInstances` — spins up a second `ATTNSDK` and confirms each instance's notification carries its own `object`, so a multi-instance host can filter cleanly.

368/368 passing.

Sources/Inbox/InboxManager.swift
adelag-attentive: Done in 24c40d7. Added an "Ordering invariant" paragraph to the `UnreadCountBox` class comment stating that `onChange` callbacks are only guaranteed in write-order under a single serialized writer, and that today's only writer is the `InboxManager` actor's `didSet` (safe because actor isolation serializes all writes). Noted explicitly so a future maintainer doesn't add a second concurrent writer without gating the callback.

usharif-attentive (approved):

PR #295, opened 2026-08-07
MSDK-312: fix root causes of concurrency & webview test flake

usharif-attentive:
## Summary

Fixes the root causes of the two chronic CI flakes in `ATTNUserIdentityTests` and `ATTNWebViewHandlerIntegrationTests` (most recently pipeline 603). Previous attempts (timeout bumps, iteration cuts, storage stubbing, log silencing, bounded parallelism) tuned parameters around the failures without removing the mechanisms; both mechanisms are now structurally eliminated. Test-only change — no SDK source is touched.

## Key Changes

**Identity concurrency tests: QoS-starved lock convoy.** `runConcurrently` drove its contended-`NSLock` stress workload through `DispatchQueue.global()` at `.default` QoS while the main thread parked in `group.wait`. A test hammering one lock is a lock convoy — wall time is `handoffs × thread-wakeup latency` — and on loaded CI VMs, wakeup latency for default-QoS threads is tens of milliseconds. Pipeline 603 captured the smoking gun: the identical workload ran in **0.002s** when driven from the (high-priority) test thread and **13.5s** through the helper, seconds apart in the same run. The helper now runs its workers at `.userInteractive`, and its timeout is documented as a deadlock detector (generous by design) rather than a performance bound.

**WebView suite: every test raced a live 0.2s teardown timer.** The shared `setUp` handler used `launchTimeoutInterval: 0.2`, so every launch armed a real teardown timer and every assertion between launch and close silently raced it; on a slow runner the timer won and tore down the webview mid-test ("expected CustomWebView after launch #2"). The shared handler now uses a timeout that cannot elapse mid-test; only the tests actually exercising the timeout build a short-timeout handler, and they anchor waits on the timeout's observable effects instead of wall-clock sleeps. Event-anchored waits are widened to 30s (they return instantly on success; CI showed real 8–9s scheduler stalls). Also: `testLaunchCreative_ShouldLoadCorrectURL` no longer uses the shared state manager (cross-test pollution risk), and the multi-launch settle window widened 0.05s → 0.5s (negative check — too short risks false-pass).

## Verification

- Full unit suite passes locally: 227 tests, 0 failures (iPhone 17 simulator, iOS 26.1, Xcode 26.x via direct `xcodebuild`).
- Both suites stress-tested locally with `-test-iterations 15 -run-tests-until-failure`: 345 test executions, 0 failures.
- **7 consecutive green CI runs** on this branch (pipelines 606–613, m4pro.medium, Xcode 26.1.1).
- CI-measured proof of mechanism (pipeline 609 timings): `testClearUser_concurrentClearAndMerge` **0.004s** (previously >15s timeout) and `testMergeAndRead_concurrentReadersAndWriters` **0.002s** (previously 13.48s) — ~4 orders of magnitude, on CI hardware.
- CircleCI's flake detector flags exactly one historically flaky test in this project — `testClearUser_concurrentClearAndMerge_doesNotCrash`, at two different line numbers across commits — confirming the historical flake is confined to what this PR fixes.

## Risk / Rollback

Low: test-target only, no public API or SDK behavior change, nothing ships to host apps. Race coverage is preserved (same iteration counts, same interleavings — only worker priority and wait budgets changed). Rollback is a plain revert.

### DIFF
--- a/Tests/TestCases/ATTNUserIdentityTests.swift
+++ b/Tests/TestCases/ATTNUserIdentityTests.swift
@@ -101,7 +101,7 @@ final class ATTNUserIdentityTests: XCTestCase {
 
     func testMergeAndRead_concurrentReadersAndWriters_doesNotCrash() {
         let identity = ATTNUserIdentity(identifiers: [ATTNIdentifierType.email: "seed@test.com"])
-        runConcurrently(iterations: 200, timeout: 15, queueLabels: ["writer", "reader"]) { i, queueIndex in
+        runConcurrently(iterations: 200, queueLabels: ["writer", "reader"]) { i, queueIndex in
             if queueIndex == 0 {
                 identity.mergeIdentifiers([ATTNIdentifierType.email: "user\(i)@test.com"])
             } else {
@@ -117,7 +117,7 @@ final class ATTNUserIdentityTests: XCTestCase {
         // with clearUser() logging outside its critical section, each os_log
         // call is still on the timed path (the dispatch block doesn't finish
         // until the log returns), and CircleCI's log capture serializes os_log
-        // at ~75ms/call, which at 200 iterations would blow the timeout.
+        // at ~75ms/call, which at 200 iterations adds ~15s of pure log latency.
         let identity = ATTNUserIdentity(
             identifiers: [:],
             visitorService: ATTNVisitorService(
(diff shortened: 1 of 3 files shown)


usharif-attentive:
@codex review

usharif-attentive:
@attentive-claude review

